PR #54, opened 2019-04-05
Run sequential tests separately

sebbonnet:
As sequential tests are not designed to be run in parallel.
In fact this was causing the build to fail intermittently as resources
created by a test where deleted by another.

Also remove E2E_TEST parameter as it no longer fit
with the test model now that the e2e tests have been split up in 2
groups. Trying to support single test run in this model seemed
convulated and has little benefit since most users would use
their IDE to run tests.

Also make the test cleanup a bit more robus by cleaning cassandra resources explicitly


cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -328,6 +329,16 @@ func Kubectl(namespace string, args ...string) (*exec.Cmd, []byte, error) {
 	return cmd, output, err
 }
 
+func CassandraDefinitions(namespace string) func() ([]v1alpha1.Cassandra, error) {
```
alangibson01: Couldn't the return type just be `([]v1alpha1.Cassandra, error)`?

sebbonnet: Yep - I initially used it to in a `gomega.Eventually` method call to wait for all resources in the namespace to be deleted, but we can't do that as in some cases we need to delete specific cassandra definition in the namespace. Adding the logic to wait until specific cassandra definition is possible, but i don't think it's necessary as these types of resources are deleted immediately.

PR #55, opened 2019-04-05
Specify resource constrainst for test pods

sebbonnet:
So tests can be run against a cluster which requires resources
requirements to be specified.

Also removes duplicated cluster output from the test failure diagnosis


sebbonnet:
This branch probably won't pass until #54 is merged

PR #58, opened 2019-04-24
Expose ginkgo options to limit resource usage

sebbonnet:
Add environment variables to Makefiles to allow specifying options for ginkgo `-nodes` and `-compilers`.
This is to better control the number of cpu assigned for compiling and running
tests in parallel, as the default values prove too much for some environments.
Defaulting both values to zero is equivalent to not specifying the options at all.


CONTRIBUTING.md
```diff
@@ -45,6 +45,8 @@ Flag | Meaning | Default
 `TEST_REGISTRY`                | The name of the docker registry where test images created via the build will be pushed| `localhost:5000`
 `DOCKER_USERNAME`              | The docker username allowed to push to the release registry | (provided as encrypted variable in `.travis.yml`)
 `DOCKER_PASSWORD`              | The password for the docker username allowed to push to the release registry | (provided as encrypted variable in `.travis.yml`)
+`GINKGO_COMPILERS`             | Ginkgo `-compilers` value to use when compiling multiple tests suite | `0`, similar to not settings the option at all
```
alangibson01:
```suggestion
`GINKGO_COMPILERS`             | Ginkgo `-compilers` value to use when compiling multiple tests suite | `0`, equivalent to not setting the option at all
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,8 @@ Flag | Meaning | Default
...
+`GINKGO_NODES`                 | Ginkgo `-nodes` value to use when running tests suite in parallel | `0`, similar to not settings the option at all
```
alangibson01:
```suggestion
`GINKGO_NODES`                 | Ginkgo `-nodes` value to use when running tests suite in parallel | `0`, equivalent to not setting the option at all
```

sebbonnet:
sounds good :+1:

PR #59, opened 2019-05-07
Add note about pre-requests to contributing doc

kragniz:


CONTRIBUTING.md
```diff
@@ -7,6 +7,18 @@ issue before making a change.  This saves everyone from wasted effort in the eve
 changes need some adjustment before they are ready for submission.
 All new code, including changes to existing code, should be tested and have a corresponding test added or updated where applicable.
 
+
+## Pre-requests
```
lw346:
```suggestion
## Prerequisites
```

CONTRIBUTING.md
```diff
@@ -7,6 +7,18 @@ issue before making a change.  This saves everyone from wasted effort in the eve
...
+
+The following must be installed on your development machine:
+
+- `docker`
+- `openjdk-8`
```
lw346: While installing `openjdk-8` will guarantee the requirement of a Java 8 JDK, we should probably make clear that other JDKs that fulfill the requirement are available.

kragniz: Sure, 
```
`openjdk-8` or another JDK
```
sound okay?

CONTRIBUTING.md
```diff
@@ -7,6 +7,18 @@ issue before making a change.  This saves everyone from wasted effort in the eve
...
 
+
+## Prerequisites
+
+The following must be installed on your development machine:
+
+- `docker`
+- `openjdk-8` or another JDK
+- `gcc` (or `build-essential` package on debian distributions)
+- `kubectl`
```
sebbonnet: We could add `golang` and `dgoss` to this list too.

kragniz: Good idea.

kragniz: Though `make setup` installs `dgoss` in the "Building and Testing" step.

sebbonnet: It does you're right, but I don't think this is correct as it requires elevated privileges (`sudo`) to install `dgoss`. I have removed it in a follow up PR #64, so I think we should add the `dgoss` requirement here

kragniz: :+1:

Issue #33, opened 2019-01-02
Ensure conformity with Kubernetes API conventions


alangibson01:
We should ensure that our definition of the Cassandra kind conforms with the conventions set out in the [Kubernetes API conventions document](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md).

Examples:
- [Defaulting](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#defaulting) - done as part of #71 
  Specifically, we ought to "fill in" any values which are defaulted when we receive a Cassandra definition.
- Don't use unsigned integers in APIs, for the reasons described in [Primitive Types](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#primitive-types)
- CRD definitions use of optional fields. As per https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required, optional fields should have their type as a Go pointer, in order to allow them to be set to nil. This allows the operator to differentiate between unset fields and default values.

**Tasks**
- [x] Rename the DC field to Datacenter and make it a string pointer - parent of #61
- [x] Make all optional fields pointer types and +omitempty - parent of #72
- [x] Add OpenAPI validation to CRD resource - parent of #82
- [x] `Pod.Memory` and `Pod.CPU` should be moved to a core.v1.ResourceRequirements field - parent of #79 
- [x] Move Pod.StorageSize to Rack - parent of #78
- [x] Consider removing CassandraSpec.UseEmptyDir in favour of rack level persistence configuration - parent of #74 
- [x] Remove the `RetentionPolicy.Enabled` flag - parent of #135
- [x] Update WIKI with the new Cassandra definition - parent of sky-uk/cassandra-operator-wiki#4, sky-uk/cassandra-operator-wiki#5
- [x] Release changelog - parent of  #191

sebbonnet:
I've detached #71 from this epic, as it is won't change our API but related to how we default and validate values

sebbonnet:
Reopened to add these:
> Remove the RetentionPolicy.Enabled flag #135
> Pod.Memory and Pod.CPU should be moved to a core.v1.ResourceRequirements field #79

sebbonnet:
All sub-tasks completed

PR #61, opened 2019-05-07
Rename the DC field to Datacenter and make it a string pointer

wallrj:
Part of #33


wallrj:
Rebased and squashed.

Issue #67, opened 2019-05-08
`make validate-licences` fails with error: `Lock inputs-digest mismatch`


wallrj:
```
make -C cassandra-operator validate-licences
...
make: Entering directory '/home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-operator'
== validate licences
set -e ;\
restricted=$(paste -s -d ',' restricted-licences.txt) ;\
projects=$(dep status -f='vendor/{{ .ProjectRoot }} ') ;\
licence-compliance-checker -L error -E -r $restricted -o vendor/github.com/ghodss/yaml=BSD-3-Clause -o vendor/github.com/golang/protobuf=BSD-3-Clause -o vendor/github.com/gogo/protobuf=BSD-3-Clause -o vendor/github.com/hashicorp/golang-lru=MPL-2.0 $projects ;\

Lock inputs-digest mismatch. This happens when Gopkg.toml is modified.
Run `dep ensure` to regenerate the inputs-digest.
input-digest mismatch
make: *** [Makefile:123: validate-licences] Error 1
make: Leaving directory '/home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-operator'
```

```
make -C cassandra-snapshot validate-licences
make: Entering directory '/home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-snapshot'
== validate licences
set -e ;\
restricted=$(paste -s -d ',' restricted-licences.txt) ;\
projects=$(dep status -f='vendor/{{ .ProjectRoot }} ') ;\
licence-compliance-checker -L error -E -r $restricted -o vendor/github.com/ghodss/yaml=BSD-3-Clause $projects ;\

Lock inputs-digest mismatch. This happens when Gopkg.toml is modified.
Run `dep ensure` to regenerate the inputs-digest.
input-digest mismatch
make: *** [Makefile:58: validate-licences] Error 1
make: Leaving directory '/home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-snapshot'
```

sebbonnet:
Can you add your `Gopkg.toml` and the output of `dep status`?
It's probably just of case of adding `go-dep` as a dependency to `validate-licenses` to ensure `dep ensure` is executed before.

sebbonnet:
Now that we have moved to go modules, I think we can probably close this issue.
@wallrj do you agree?

wallrj:
Yep, this now works for me

```
[...]

Issue #66, opened 2019-05-08
The CI system is slow to report static check errors (go fmt, go vet, etc)


wallrj:
In https://github.com/sky-uk/cassandra-operator/pull/61, for example, I missed some documentation for some new public functions, but it took > 20 minutes for this to be reported because the E2E tests had to run first.

Additionally, it's not clear which `make` target i can use locally to run these static checks before committing and pushin.

For a start I propose:
 * [x]  `make verify-all` which I can use to  run these fast checks locally (fmt vet lint validate-licences) (https://github.com/sky-uk/cassandra-operator/pull/68)
*  [ ] Add a Travis build stage called (check) to with a job called `verify` to run these fast checks before the e2e: https://docs.travis-ci.com/user/build-stages/ . 
* [ ] And another job called `test` in the `check` stage which builds the binaries and then runs the tests.

sebbonnet:
I like the idea of splitting up the build so we get an earlier feedback for the static checks.
We would also need to split up the dind cluster creation from the setup as this takes ~10mins.

Given that jobs within a stage run in parallel, I think we need multiple stages rather than multiple jobs. Perhaps `check-style` to run the static checks and `check` where we would create the dind cluster before running the test.

PR #68, opened 2019-05-08
WIP: Add a make verify-all target

wallrj:
Part of #66 

This allows me to quickly run the static checks on my laptop before committing.

I tried also running validate-licences as part of `verify` but it failed (on my laptop). See: https://github.com/sky-uk/cassandra-operator/issues/67


wallrj:
Rebased and squashed.

cassandra-operator/Makefile
```diff
@@ -29,7 +29,9 @@ build: go-build go-install docker-image
 
 install: build push-image
 
-check: fmt vet lint validate-licences test deploy-operator integration-test e2e-test
+verify: fmt vet lint
```
sebbonnet: I'd rather call this `check-style`, as `verify` as a similar meaning than `check`.
Having the target `check` then calling `check-style` fits better.

Makefile
```diff
@@ -81,3 +81,10 @@ release-all:
 	$(MAKE) -C cassandra-bootstrapper release
 	$(MAKE) -C cassandra-snapshot release
 	$(MAKE) -C cassandra-operator release
+
+verify-all:
```
sebbonnet: I'd rather call this `check-style-all` to carry a better meaning.
We also need a `check-style` in that top level Makefile which would depend on `check-style-all`, so that the same target is used in all Makefile.

We do this already for install, clean and release -e.g.
```
release: release-all
```

wallrj:
Thanks for the review @sebbonnet . I think I've addresses all your comments. Back to you.

PR #83, opened 2019-05-08
WIP: Run check-style and e2e tests in parallel

wallrj:
Closes: #66


wallrj:
It seems that Travis doesn't yet support failing fast when one of many parallel job fails. See https://github.com/travis-ci/travis-ci/issues/2062

There's a `fast_finish` option, but it only applies to jobs that have been marked as `allowed_failure`.

Perhaps we should consider a different CI system.

This change at least allows you to quickly see whether the style checks have failed, but it won't fail the build until the e2e tests have also finished.

.travis.yml
```diff
@@ -8,26 +8,36 @@ go:
...
-script: make travis
-
-deploy:
-- provider: script
-  skip_cleanup: true
-  script: make release
-  on:
-    tags: true
+matrix:
+  fast_finish: true
```
wallrj: This should probably be removed as it only affects jobs marked as `allow_failure`.

.travis.yml
```diff
@@ -8,26 +8,36 @@ go:
...
+
+jobs:
+  include:
+    - stage: test
+      name: check-style
+      install: make setup-all
```
wallrj: I wonder if  we should create a Docker image containing all the build and test dependencies that we need.
Then our jobs wouldn't have to install all these tools repeatedly.
And on our laptops we wouldn't need to install these tools.

.travis.yml
```diff
@@ -8,26 +8,36 @@ go:
...
+      name: e2e
+      before_install:
+      - sudo apt-get update
+      - sudo apt-get install -y curl
+      - sudo curl -L https://storage.googleapis.com/kubernetes-release/release/v1.10.13/bin/linux/amd64/kubectl -o /usr/local/bin/kubectl
+      - sudo chmod +x  /usr/local/bin/kubectl
+      - sudo curl -L https://github.com/aelsabbahy/goss/releases/download/v0.3.5/goss-linux-amd64 -o /usr/local/bin/goss
+      - sudo chmod +rx /usr/local/bin/goss
+      - sudo curl -L https://raw.githubusercontent.com/aelsabbahy/goss/v0.3.5/extras/dgoss/dgoss -o /usr/local/bin/dgoss
+      - sudo chmod +rx /usr/local/bin/dgoss
```
wallrj: See above.
Or perhaps all these installation commands could be moved to a `hack/install-dependencies.sh` script.

.travis.yml
```diff
@@ -8,26 +8,36 @@ go:
...
+      install: make setup
+      script: make travis
```
wallrj: It'd be nice if there were a  way to run the e2e tests separately from the unit tests.
Is there?

.travis.yml
```diff
@@ -8,26 +8,36 @@ go:
...
+
+    - stage: deploy
+      deploy:
+        skip_cleanup: true
+        script: make release
+        on:
+          tags: true
```
wallrj: I'm not sure if this is the correct way to run deploy scripts when using build stages.
The example, makes deploy, just another build job, with script: https://docs.travis-ci.com/user/build-stages/

sebbonnet:
I think we should stick to a simple workflow as much as possible: e.g. `setup install check`.
But at the same time give feedback as early as possible: run check-style, unit tests before running e2e tests. Creating the dind cluster is both so time consuming and different than anything else, that I think it deserves its own target.

So for travis.yml I would have a single job that invokes the targets sequentially - in fact we don't even need a travis job/stage, we can simply have multiple commands invoked in the travis `script` task :
```
script: make setup install dind check
```

- `make setup`: doesn't need changing, except at the top-level Makefile we should remove `recreate-dind-cluster`
- `make install`: builds and installs the artifacts for each module: jar + docker image OR go binary + docker image.
   - install for "java modules" already does what we want: 
     - it compiles, runs the unit test and packages the artifact and builds and push the docker image. 
     - it would eventually run check-style, probably via gradle, so I think we don't need `check-style` as a separate target here
     - it would eventually run the licence compliance check, again probably via gradle
   - install for "go modules" is where we need to change things a bit
     - build: I would make it run `check-style` and `test` (maybe better named as `unit-test`) before building the docker image. I think it would make sense to validate the licence compliance here too. 
     - install: I think that's ok, except I would have `go-install` here rather than in `build` 
- `make dind`: this would recreate the dind cluster
- `make check`: to me this should be used to run tests against the artifacts produced
   - check for "java modules" is pretty much unchanged and would simply run dgoss
   - check for "go modules" would run integration and e2e tests as well as setting up k8 resources

I think the travis target no longer makes any sense, so we should rename that to `all` or similar.

.travis.yml
alangibson01: That'd be a good idea, as we'd be able to exactly reproduce the CI build when we run locally.

Can we do this in Travis? I know that with CircleCI you can choose a Docker container which your build runs in, but I couldn't find anything that offers the same ability in Travis.

.travis.yml
alangibson01: Within the makefiles, the unit tests are run by a separate target from the E2E tests, with the unit tests coming first. Is your concern being able to see from the Travis interface at which step a test failure happened?

wallrj:
Thanks @sebbonnet I'm going to try and implement your suggested CI sequence in #99

.travis.yml
wallrj: My mistake. I'd missed the `make -C cassandra-operator test` target  when I wrote that comment.

.travis.yml
wallrj: Done, in #99

.travis.yml
wallrj: Hmm, actually it'd require changing to the travis VM runner, which is slower....but might be worth it, to avoid re-installing the dependencies on every CI run.

Issue #81, opened 2019-05-08
Consider making Snapshot a top-level resource


kragniz:
Consider making Snapshot a top-level resource, operating more like a kubernetes `Job`.

A `SnapshotSchedule` resource could then layer on top to periodically create Snapshots

This would allow more flexibility in managing snapshotting, for example different schedules for individual keyspaces.

Also allow better error reporting via `Snapshot.Status`.

alangibson01:
Could you sketch out how you envisage the API looking? I'm not sure I understand what you mean about `SnapshotSchedule` layering on top.

Another thought is that if we split snapshots out from the cluster spec, we'd need to account for the possibility of a cluster being deleted but snapshots which reference it still existing.

wallrj:
@alangibson01 to be honest, @kragniz and I were doing a lot of hand waving at this stage, but I think  we were being influenced by other Backup APIs we've seen. For example:
* https://github.com/heptio/velero/blob/master/pkg/apis/velero/v1/backup.go
* https://github.com/heptio/velero/blob/master/pkg/apis/velero/v1/schedule.go

And: 
 * https://github.com/coreos/etcd-operator/blob/master/doc/user/walkthrough/backup-operator.md

A Snapshot existing without a cluster may be a useful thing, if the cluster has been accidentally deleted, but the data still exists on PVs.

wallrj:
@kragniz pointed out this Cassandra backup API https://github.com/instaclustr/cassandra-operator/blob/master/doc/backup_restore.md  which we could examine for ideas.

sebbonnet:
Closing as explained here: https://github.com/sky-uk/cassandra-operator/pull/84#discussion_r285077126

Issue #78, opened 2019-05-08
Move Pod.StorageSize to Rack


kragniz:
This should be located alongside `Rack.StorageClass`.

UDPDATE: @sebbonnet 
We're taking the opportunity the completely overhaul the way we define and handle rack storage as described here: https://github.com/sky-uk/cassandra-operator/issues/78#issuecomment-528850121
- [x] Allow multiple rack storage #184, #190 
- [x] Allow updates and removal of emptyDir storage #192  

Part of #33.

alangibson01:
This would mean you could have different storage sizes on different racks, which I don't think you would want on a Cassandra cluster. The split of the token range across nodes would need to account for the different amounts of storage available on each node, which is not a scenario we see as being likely.

wallrj:
@alangibson01 That makes sense. For comparison other C* operators :
 * Cluster level: https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/db2689105fa932a4ebc8a0cac49ddf438f4c00d7/pkg/apis/db/v1alpha1/cassandracluster_types.go#L620
 * Rack level: https://github.com/rook/rook/blob/release-1.0/Documentation/cassandra-cluster-crd.md#sample
 * Rack level: https://github.com/scylladb/scylla-operator/blob/master/docs/scylla_cluster_crd.md#sample
 * DC level: https://github.com/instaclustr/cassandra-operator/blob/master/examples/common/test.yaml#L26

sebbonnet:
I can see the argument to move both storage class and size together.
As storage class cannot be moved to pod for the reason explained [here](https://github.com/sky-uk/cassandra-operator/pull/84#discussion_r283710410 ), the only option is to move both properties to Racks.

I get your point @alangibson01, but I think this decision would be up to the users.
Although we would recommend using the same disk type and size across all nodes, I don't think we want to prevent the users from doing it.

sebbonnet:
Having all storage attributes together (zone, store class and size) makes sense and is arguably a better design. With our current structure the users could already have different underlying disk types for each rack, which is obviously a bad idea, but not something we should prevent.
Happy to reopen

sebbonnet:
I think we should change our API to be closer to the Kubernetes API, so that users can be in full control of how they specify storage properties and so our API naturally evolves alongside Kubernetes. 

[...]

PR #84, opened 2019-05-09
WIP: Proposed improved API as YAML

wallrj:
Part of #33

### DIFF
--- a/examples/cassandra1.yaml
+++ b/examples/cassandra1.yaml
@@ -12,11 +12,16 @@ spec:
     persistence:
       storageClass: "local-ssh"
       size: "100Gi"
+    resources:
+      requests:
+        cpu: "500m"
+        memory: "1Gi"
+      limits:
+        cpu: "1"
+        memory: "2Gi"
   pod:
     bootstrapperImage: "skyuk/cassandra-bootstrapper:latest"
     image: "cassandra:3.11"
-    memory: "1Gi"
-    cpu: "1"
     livenessProbe:
       failureThreshold: 1
       initialDelaySeconds: 10
--- a/examples/cassandra1.yaml
+++ b/examples/cassandra1.yaml
@@ -34,15 +34,36 @@ spec:
       limits:
         cpu: "1"
         memory: "2Gi"
-  snapshot:
+---
+apiVersion: "core.sky.uk/v1alpha1"
+kind: "CassandraSnapshotSchedule"
+metadata:
+  name: "cluster-1-daily"
+  namespace: "example.com"
+spec:
+  schedule: "0 4 * * *"
+  cleanupSchedule: "0 4 * * *"
+  template:
+    clusterName: "cluster-1"
     image: "skyuk/cassandra-snapshot:latest"
-    schedule: "0 4 * * *"
     keyspaces:
     - "mykeyspace"
     - "system_auth"
     timeoutSeconds: 30
     retentionPolicy:
-      enabled: true
       retentionPeriodDays: 60
-      cleanupSchedule: "0 4 * * *"
-      cleanupTimeoutSeconds: "600"
+---
+apiVersion: "core.sky.uk/v1alpha1"
+kind: "CassandraSnapshot"
+metadata:
+  name: "cluster-1-ondemand-snapshot-1"
+  namespace: "example.com"
+spec:
+  clusterName: "cluster-1"
+  image: "skyuk/cassandra-snapshot:latest"
+  keyspaces:
+  - "mykeyspace"
+  - "system_auth"
+  timeoutSeconds: 30
+  retentionPolicy:
+    retentionPeriodDays: 10
--- a/examples/cassandra1.yaml
+++ b/examples/cassandra1.yaml
@@ -12,13 +12,6 @@ spec:
     persistence:
       storageClass: "local-ssh"
       size: "100Gi"
-    resources:
-      requests:
-        cpu: "500m"
-        memory: "1Gi"
-      limits:
-        cpu: "1"
-        memory: "2Gi"
   pod:
     bootstrapperImage: "skyuk/cassandra-bootstrapper:latest"
     image: "cassandra:3.11"
@@ -34,6 +27,13 @@ spec:
       periodSeconds: 10
       successThreshold: 3
       timeoutSeconds: 30
+    resources:
+      requests:
+        cpu: "500m"
+        memory: "1Gi"
+      limits:
+        cpu: "1"
+        memory: "2Gi"
   snapshot:
     image: "skyuk/cassandra-snapshot:latest"
     schedule: "0 4 * * *"
--- a/examples/cassandra1.yaml
+++ b/examples/cassandra1.yaml
@@ -9,12 +9,12 @@ spec:
   - name: "region-1-zone-a"
     zone: "zone-a"
     replicas: 3
-    persistence:
-      storageClass: "local-ssh"
-      size: "100Gi"
   pod:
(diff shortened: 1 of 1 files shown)


examples/cassandra1.yaml
```diff
@@ -0,0 +1,69 @@
...
+  datacenter: "region-1"
+  racks:
+  - name: "region-1-zone-a"
+    zone: "zone-a"
+    replicas: 3
+  nodePodConfiguration:
+    bootstrapperImage: "skyuk/cassandra-bootstrapper:latest"
+    image: "cassandra:3.11"
+    persistence:
+      storageClass: "local-ssd"
```
sebbonnet: A storage class can be either global or zone specific.
In our case we want it to be zone specific, so we can create a Statefulset (Rack) per zone that would use PVs in the same zone.
The concept of Rack is essential to provide additional resilience to Cassandra and is used internally to decide which replicas should own the data when `NetworkTopologyStrategy` is used:
```
NetworkTopologyStrategy attempts to place replicas on distinct racks because nodes in the same rack (or similar physical grouping) often fail at the same time due to power, cooling, or network issues.
```

So basically a storageClass is best placed inside the rack struct.

examples/cassandra1.yaml
```diff
@@ -0,0 +1,69 @@
...
+metadata:
+  name: "cluster-1"
+  namespace: "example.com"
+spec:
+  datacenter: "region-1"
+  racks:
+  - name: "region-1-zone-a"
+    zone: "zone-a"
+    replicas: 3
+  nodePodConfiguration:
```
sebbonnet: I prefer the previous term `pods` as it better align with `racks`. 
We should be consistent within the same config

examples/cassandra1.yaml
```diff
@@ -0,0 +1,69 @@
...
+      timeoutSeconds: 30
+    resources:
+      requests:
+        cpu: "500m"
+        memory: "1Gi"
+      limits:
+        cpu: "1"
+        memory: "2Gi"
+---
+apiVersion: "core.sky.uk/v1alpha1"
```
sebbonnet: What is the value of splitting Snapshot and its Schedule out of the Cassandra spec?
I'm not sure having a SnapshotSchedule or Snapshot resource on their own with no corresponding Cassandra cluster provides any benefits.
On the other hand the snapshot `data` that is created by the snapshot job can continue to exist without a cluster. In fact when we delete a Cassandra resource all the associated K8 resources, we don't delete the PVs so that the data (snapshot or live) can be preserved. The snapshot data is still available regardless of whether Cassandra snapshot jobs or Cassandra resource exists. 
Similarly when we implement external storage for snapshots in the future (e.g. AWS S3), we could complete delete all K8 resources while preserving the snapshot data.

examples/cassandra1.yaml
```diff
@@ -0,0 +1,69 @@
...
+  schedule: "0 4 * * *"
+  cleanupSchedule: "0 4 * * *"
+  template:
+    clusterName: "cluster-1"
+    image: "skyuk/cassandra-snapshot:latest"
+    keyspaces:
+    - "mykeyspace"
+    - "system_auth"
+    timeoutSeconds: 30
+    retentionPolicy:
```
sebbonnet: Why removing the `enabled` property? I thought it would be handy to make it easier to toggle on / off the cleanup.

examples/cassandra1.yaml
wallrj: Ok. But worth considering that the name `Pod` may make users think that this field accepts a full PodSpec or a PodTemplate.
It contains a small number of `C* node` configuration options.

examples/cassandra1.yaml
wallrj: This will allow different C* nodes to have different storage classes. In https://github.com/sky-uk/cassandra-operator/issues/78 we argue that all C* nodes should have the same storage capacity. Should they not also have the same storage throughput etc?

Different cloud zones may not all support the same storage types, so it may be necessary to have different Racks (in different Cloud availability zones) use different storage classes.
But would you not ensure that all Cassandra Racks run in zones with identical properties?

examples/cassandra1.yaml
wallrj: I'm not totally sure of this change either, but here are some points:
* A Snapshot resource gives us a place to record a Snapshot status.
* A snapshot controller can be tightly restricted to operate only on Snapshot resource, not Cassandra resources.
* A Snapshot resource (stored on S3)  could be referred to by a controller which is able to create / restore a Cassandra database.

examples/cassandra1.yaml
wallrj: I thought it simpler to omit the  `cleanupSchedule`. Controller would interpret `nil` as no cleanup.

wallrj: We could also drop the Cron syntax, which I never remember, and using something clearer.

examples/cassandra1.yaml
sebbonnet: C* nodes should have similar storage properties. 
If there was a way we could _guarantee_ that pods are evenly distributed across availability zones with a single storage class, then we could have both storage class and size at the pod configuration level - therefore ensuring that the storage properties are the same across C* nodes.
However when using Statefulsets, Kubernetes would only try spreading pods across zones on a best effort basis - there is no way we can influence that. That is not good enough as we must guarantee that racks are in different physical grouping, as Cassandra rack model expects.   

Are you trying to argue for a persistence entity that has both storage class and size?

examples/cassandra1.yaml
sebbonnet: I see your point - perhaps just `node` then?

examples/cassandra1.yaml
sebbonnet: These are all good points, although I think it would be nicer to set the statuses of both snapshop and snapshot cleanup on the Cassandra resources - instead of having to check 2 different resources. We could even keep a history of snapshots/cleanup runs, must that's getting ahead of myself :)

examples/cassandra1.yaml
sebbonnet: OK about removing the `enabled` - I don't feel strongly about it

sebbonnet: haha - true about the cron syntax, I'm the same.
But I also think we should follow the kubernetes spec layout as much as possible to make it easier to correlate with the k8 resource that would be eventually created.

examples/cassandra1.yaml
wallrj: Now that I'm more familiar with the project I don't think this change is really necessary.
It'll just cause a load of code churn.

examples/cassandra1.yaml
wallrj: Yes, I think storage class and size belong together. See https://github.com/sky-uk/cassandra-operator/issues/78#issuecomment-499939352 for some examples of other Cassandra operators that follow that pattern.

examples/cassandra1.yaml
wallrj: Agreed.

examples/cassandra1.yaml
wallrj: Ok. I'll abandon this proposed change for now. Although worth looking at the examples of top level backups and snapshots in https://github.com/sky-uk/cassandra-operator/issues/81#issuecomment-490542815

examples/cassandra1.yaml
sebbonnet: I agree that having both storage class and size together would make sense. But has storage class cannot be moved to pod for the reason explained above, the only other option would be to move both properties to Racks

Issue #74, opened 2019-05-08
Consider removing CassandraSpec.UseEmptyDir in favour of rack level persistence configuration


kragniz:
Part of #33.

alangibson01:
Getting rid of `UseEmptyDir` would definitely make things cleaner.

We already have a `StorageClass` field on the rack - how would this new `Persistence` field differ?

sebbonnet:
Closed via #185

PR #85, opened 2019-05-09
WIP: Persistence configuration (Design)

wallrj:
Closes: #74 
Part of: #33

### DIFF
--- a/cassandra-operator/CHANGELOG.md
+++ b/cassandra-operator/CHANGELOG.md
@@ -2,25 +2,25 @@
 
 ## 0.70.0-alpha
 - [ANNOUNCEMENT] This is the first release available to the general public
-  
+
   The operator is alpha-status software and can be used in development environments.
-  It is not yet recommended for use in production environments.      
-  
+  It is not yet recommended for use in production environments.
+
 - [FEATURE] New properties to specify docker images for snapshot and bootstrapping
-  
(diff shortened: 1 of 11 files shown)


alangibson01:
I think we need to keep the StorageClass on the rack, as each rack requires storage which is local to that rack. For example, if your Kubernetes cluster is in AWS across multiple AZs, for redundancy purposes you'd want each of your racks to be in a separate AZ, and you'd most likely have a number of different storage classes, each of which specifically targets a given AZ.

I'm still not convinced that Snapshot needs to live outside the Cassandra resource. The lifecycles of these two would be very tightly coupled - with no cluster to take a snapshot of, the snapshot would have no purpose. Is there a particular benefit you can see of having separate snapshots?

wallrj:
I may be misunderstanding the problem.
I think that if you specify `WaitForFirstConsumer` https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode, then Kubernetes won't create a PV until the Pod has been scheduled. 
And assuming that the PodTemplate of each StatefulSet restricts its pods to a single AZ, then the PV should always be available.

alangibson01:
@wallrj Sorry for the delay in getting back to you, I've been on holiday for the last week. I think the rack/storage class discussion we started here is the same as the one going on in #84 [here](https://github.com/sky-uk/cassandra-operator/pull/84#discussion_r282879957).

Issue #60, opened 2019-05-07
Migrate from dep to Go modules


kragniz:
The project currently uses `dep` to manage go dependencies. Go modules are generally more pleasant to use, but require Go 1.11.

Migrate from dep to Go modules at some point after moving to Go 1.11 or newer.

sebbonnet:
I haven't used go modules yet, but the fact that dep management is integrated in standard `go build` and `go test` commands sound like a good thing.
We would need to find replacement for `dep status` too which we use to pass the list of vendor dependencies to the tool checking licence compliance.

Travis CI already builds using go 1.11

sebbonnet:
Work done in #86

PR #87, opened 2019-05-10
Add a webhook for validation of Cassandra resources

wallrj:
Required by #71 

This has taken a while, but I think I finally got a bare bones validating webhook ready for a review.

There is no complex validation yet, but the webhook does, at least, validate that the submitted json can be unmarshalled to a Cassandra object and it will report a validation error if the submitted yaml contains unexpected fields.

The webhook uses https://github.com/openshift/generic-admission-server 
It works by registering its self as an aggregated API server.
And the webhook is therefore defined as a request to the Kubernetes API rather than to a separate service.
This is the same technique used in cert-manager and it makes it simpler to set up the mutual authentication between the webhook service and the api server.

The webhook is currently in a separate sub-project, because I couldn't get dep to resolve all the dependencies when I introduced k8s.io/apiserver dependency.

The complex validation functions will live in the cassandra-operator sub-project so that i can re-use them for controller level validation (for cases where the webhook has not been deployed).

I'll follow up with branches for documentation and further validation.


Makefile
```diff
@@ -8,6 +8,7 @@ USE_MOCK ?= true
 FAKE_CASSANDRA_IMAGE ?= $(TEST_REGISTRY)/fake-cassandra:v$(gitRev)
 CASSANDRA_BOOTSTRAPPER_IMAGE ?= $(TEST_REGISTRY)/cassandra-bootstrapper:v$(gitRev)
 CASSANDRA_SNAPSHOT_IMAGE ?= $(TEST_REGISTRY)/cassandra-snapshot:v$(gitRev)
+CASSANDRA_WEBHOOK_IMAGE ?= $(TEST_REGISTRY)/cassandra-webhook:v$(gitRev)
```
wallrj: Unused.

cassandra-webhook/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	done;
+
+ensure-build-dir-exists:
+	mkdir -p $(buildDir)
+
+push-image: docker-image common-docker-push
+
+docker-image: go-build common-docker-build
+
+release: common-docker-release
```
wallrj: I haven't tested this.

cassandra-webhook/deploy.sh
```diff
@@ -0,0 +1,101 @@
...
+: ${NAMESPACE?${usage}}
+
+repoDir="$(git rev-parse --show-toplevel)"
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+scriptPath="${scriptDir}/$(basename ${BASH_SOURCE[0]})"
+templatesDir="${scriptDir}/kubernetes-resources"
+resourcesDir="${scriptDir}/build/deploy.sh.files"
+name="cassandra-operator-webhook"
+
+export CA_CERT="$(docker exec kube-master cat /etc/kubernetes/pki/ca.crt | base64 | tr -d '\n')"
```
wallrj: There might be a better way to get this. Perhaps from `kubectl config`.

cassandra-webhook/deploy.sh
```diff
@@ -0,0 +1,101 @@
...
+EOF
+
+    kubectl apply -f manifests/csr.yaml
+
+    kubectl certificate approve "${fqdn}"
+
+    while ! test -s pki/server.crt; do
+        kubectl get csr "${fqdn}" -o jsonpath='{.status.certificate}' \
+            | base64 --decode > pki/server.crt
+    done
```
wallrj: This needs a timeout and delay.

cassandra-webhook/deploy.sh
```diff
@@ -0,0 +1,101 @@
...
+    pushd "${templatesDir}"
+    find . -type f -iname '*.yaml' | while read relPath; do
+        envsubst < $relPath > "${resourcesDir}/manifests/${relPath}"
+    done
+    popd
+}
+
+function deploy() {
+    kubectl apply -f ${resourcesDir}/manifests
+    retry kubectl get --raw /apis/admission.core.sky.uk/v1beta1
```
wallrj: This only checks that the aggregate API server is ready.
Not necessarily that the API server is going to call it for webhook events.

cassandra-webhook/examples/cassandra1.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: core.sky.uk/v1alpha1
+kind: Cassandra
+metadata:
+  name: mycluster
+spec:
+  foreignFieldName: foo
```
wallrj: Rename this file and explain its intent.

cassandra-webhook/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+build: go-build go-install docker-image
+
+install: build push-image
+
+deploy: install
+	IMAGE=$(IMAGE_TO_DEPLOY) CONTEXT=$(KUBE_CONTEXT) NAMESPACE=$(NAMESPACE) $(projectDir)/deploy.sh
+
+check: deploy
+	! kubectl apply -f examples/cassandra1.yaml
```
wallrj: Maybe write a proper Ginkgo test for this.

cassandra-webhook/Makefile
```diff
@@ -0,0 +1,67 @@
...
+
+push-image: docker-image common-docker-push
+
+docker-image: go-build common-docker-build
+
+release: common-docker-release
+
+setup:
+	@echo "== setup"
+	go get -u github.com/cloudflare/cfssl/cmd/cfssl
```
sebbonnet: we should follow the tools.go best practice to track tools dependencies - see https://github.com/golang/go/issues/25922 
We've done this recently in #60

cassandra-webhook/Makefile
```diff
@@ -0,0 +1,67 @@
...
+	go get -u github.com/cloudflare/cfssl/cmd/cfssljson
+
+clean: common-docker-clean
+	@echo "== clean"
+	kubectl delete -f $(buildDir)/deploy.sh.files/manifests || true
```
sebbonnet: we should use the KUBE_CONTEXT variable to make it possible to run this build against custom contexst

cassandra-webhook/deploy.sh
sebbonnet: We should not assume that we are running against a DIND cluster, as this would limit our options in the future, should we wish to move to say KIND.
We should try using the kubernetes API as much as possible. Using kubeconfig sounds a better option than `docker exec`

cassandra-webhook/deploy.sh
```diff
@@ -0,0 +1,101 @@
...
+name="cassandra-operator-webhook"
+
+export CA_CERT="$(docker exec kube-master cat /etc/kubernetes/pki/ca.crt | base64 | tr -d '\n')"
+
+source "${repoDir}/hack/libdeploy.sh"
+
+function create_certificates() {
+    local fqdn="${name}.${NAMESPACE}.svc"
+    pushd "${resourcesDir}"
+    # See https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/
```
sebbonnet: What is the value of signing the certificate via kubernetes given that our dind cluster is setup with `insecure-skip-tls-verify`? 
Would't generating a self-signed certificate using open-ssl like in this example be enough? See https://github.com/stackrox/admission-controller-webhook-demo/blob/master/deployment/generate-keys.sh

cassandra-webhook/deploy.sh
sebbonnet: We should probably allow `CA_CERT` to be specified, so that users who want to run these tests against their cluster can provide their own certificate.
If `CA_CERT` is provided we could then skip the entire certificate generation

cassandra-webhook/deploy.sh
```diff
@@ -0,0 +1,101 @@
...
+  groups:
+  - system:authenticated
+  request: $(cat "pki/server.csr" | base64 | tr -d '\n')
+  usages:
+  - digital signature
+  - key encipherment
+  - server auth
+EOF
+
+    kubectl apply -f manifests/csr.yaml
```
sebbonnet: We should provide a KUBE_CONTEXT to all `kubectl` commands

cassandra-webhook/kubernetes-resources/validating-webhook.yaml
```diff
@@ -0,0 +1,25 @@
...
+          - v1alpha1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - cassandras
+    failurePolicy: Fail
+    clientConfig:
+      service:
+        name: kubernetes
```
sebbonnet: shouldn't that service name match the service you defined above and so be called `cassandra-operator-webhook`?
Same comment for the namespace

cassandra-webhook/kubernetes-resources/validating-webhook.yaml
```diff
@@ -0,0 +1,25 @@
...
+        namespace: default
+        path: /apis/admission.core.sky.uk/v1beta1/cassandras
+      caBundle: $CA_CERT
```
sebbonnet: it would be nice to have a namespaceSelector so users can easily toggle validation for a given Cassandra definition

cassandra-webhook/kubernetes-resources/cassandra-operator-crd.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
sebbonnet: We already have this resource in cassandra-operator. We should use it rather than duplicate it

cassandra-webhook/kubernetes-resources/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: cassandra-operator-webhook
+        deployment: cassandra-operator-webhook
+    spec:
+      containers:
+      - env:
+        - name: POD_NAMESPACE
```
sebbonnet: I'd rather name this NAMESPACE instead of POD_NAMESPACE and renamed the webhook namespace placeholder to WEBHOOK_NAMESPACE, as we already use NAMESPACE in the operator

cassandra-webhook/kubernetes-resources/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+  name: cassandra-operator-webhook
+  namespace: $NAMESPACE
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cassandra-operator-webhook
+      deployment: cassandra-operator-webhook
+  strategy:
+    type: Recreate
```
sebbonnet: I suppose there is no need to have a non-disruptive deployment for this, so Recreate rather than RollingUpdate is fine

cassandra-webhook/kubernetes-resources/rbac.yaml
```diff
@@ -0,0 +1,36 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cassandra-operator-webhook:auth-delegator
+  labels:
+    app: cassandra-operator-webhook
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:auth-delegator
```
sebbonnet: Why is this used for?

cassandra-webhook/kubernetes-resources/validating-webhook.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
```
sebbonnet: shouldn't this be a `MutatingWebhookConfiguration`, so we can apply default value to fields?

cassandra-webhook/kubernetes-resources/validating-webhook.yaml
sebbonnet: Better call it CA_BUNDLE, so it matches the field name

cassandra-webhook/kubernetes-resources/validating-webhook.yaml
sebbonnet: isn't the intention to use one webhook that would act both as a Validating and Mutating webhook?

cassandra-webhook/kubernetes-resources/apiservice.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: apiregistration.k8s.io/v1beta1
```
sebbonnet: Im not familiar with this.
Could you explain what the role of this component is?

cassandra-webhook/Makefile
wallrj: Sorry, I missed that, but worth considering whether it might be simpler to rely on the default kube context `kubectl config use-context`

cassandra-webhook/deploy.sh
wallrj: My understanding is that the `deploy.sh` scripts are intended only for deploying clusters in our tests.
But if a user wanted to run use the webhook, and run these tests against their own cluster in order to verify their installation, they won't be using this `deploy.sh` script, right?

`make -C cassandra-webook check`  will check  that invalid `Cassandra` resources are not accepted.

In which case, I should probably remove the dependency of check on deploy.

sebbonnet: You're correct - `deploy.sh` scripts are only intended to be used for tests.
However I don't think we can exclude that contributors may want to test their changes against their local cluster rather than DinD, before submitting a PR. 

I think your `check` dependency is fine.

cassandra-webhook/Makefile
sebbonnet: No worries.
I'd rather be explicit and specify the context.
I personally find that when working on multiple clusters it's easy to lose track on what the current context is.

sebbonnet:
Closing in favour of #132, as we would want to implement a simple http server based webhook rather than and extended api-server as implemented here. 
More details on that decision can be found here: https://github.com/sky-uk/cassandra-operator/pull/157

PR #88, opened 2019-05-16
types: add omitempty for optional fields

kragniz:
Including lists for consistency with kubernetes core apis.

Co-authored-by: <richard.wall@jetstack.io>

### DIFF
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
@@ -64,9 +64,9 @@ type Pod struct {
 	Memory      resource.Quantity `json:"memory"`
 	CPU         resource.Quantity `json:"cpu"`
 	// +optional
-	LivenessProbe *Probe `json:"livenessProbe"`
+	LivenessProbe *Probe `json:"livenessProbe,omitempty"`
 	// +optional
-	ReadinessProbe *Probe `json:"readinessProbe"`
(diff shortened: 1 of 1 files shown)


cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -34,7 +34,8 @@ type Cassandra struct {
 type CassandraSpec struct {
 	// +optional
 	Datacenter *string `json:"datacenter,omitempty"`
-	Racks      []Rack  `json:"racks"`
+	// +optional
+	Racks []Rack `json:"racks,omitempty"`
```
wallrj: I'm  unclear whether `// +optional` is correct here.
It's a required field, but without any validation, it is possible to omit this field.

For example, `Pod.Spec.Containers` is a required list and has neither `// + optional` nor `omitempty`: https://github.com/kubernetes/api/blob/master/core/v1/types.go

Where as `CertificateSigningRequestSpec.Usages` is not marked `// +optional` and yet  marked `omitempty`

sebbonnet: Racks should be required.
We already have validation for it in: https://github.com/sky-uk/cassandra-operator/blob/v0.70.1-alpha/cassandra-operator/pkg/cluster/cluster.go#L152

sebbonnet:
Part of #33

kragniz:
@sebbonnet I've removed rack optionality

PR #93, opened 2019-05-17
Use pointers for optional fields

kragniz:
A series of patches to move optional API fields to use pointers.

Part of #33.

### DIFF
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
@@ -36,8 +36,8 @@ type CassandraSpec struct {
 	Datacenter *string `json:"datacenter,omitempty"`
 	Racks      []Rack  `json:"racks"`
 	// +optional
-	UseEmptyDir bool `json:"useEmptyDir"`
-	Pod         Pod  `json:"pod"`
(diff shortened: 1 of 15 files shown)


cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -57,9 +64,9 @@ type Probe struct {
 
 type Pod struct {
 	// +optional
-	BootstrapperImage string `json:"bootstrapperImage"`
+	BootstrapperImage *string `json:"bootstrapperImage,omitempty"`
 	// +optional
-	Image       string            `json:"image"`
+	Image       *string           `json:"image"`
```
sebbonnet: shouldn't this use `omitempty` too?

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -839,7 +842,8 @@ var _ = Describe("creation of snapshot job", func() {
 	})
 
 	It("should create a cronjob which pod is using the specified snapshot image", func() {
-		clusterDef.Spec.Snapshot.Image = "somerepo/snapshot:v1"
+		img := "somerepo/snapshot:v1"
+		clusterDef.Spec.Snapshot.Image = &img
```
sebbonnet: could have used ptr.String, but that's picky

PR #94, opened 2019-05-17
Generate CRD using controller-tools

kragniz:
This adds validation using openapi schema.


kragniz:
WIP: the `versions` field this generates isn't compatible with 1.10, we need to move the schema to `Spec.Validation`

cassandra-operator/hack/verify-crds.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cleanup() {
+  rm -rf "${output}"
+}
+trap "cleanup" EXIT SIGINT
+
+tmp="${output}/cassandra-operator"
+
+
+rsync -avvL "${REPO_ROOT}"/ "${tmp}" >/dev/null
```
wallrj: Need to install rsync in the travis VM and add it to the make check-tools target.

cassandra-operator/hack/verify-crds.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "diffing against freshly generated crd (${tmp})"
+ret=0
+diff -Naupr "${REPO_ROOT}/kubernetes-resources/cassandra-operator-crd.yml" "${tmp}/kubernetes-resources/cassandra-operator-crd.yml" || ret=$?
+if [[ $ret -eq 0 ]]
+then
+  echo "${REPO_ROOT} up to date."
+else
+  echo "${REPO_ROOT} is out of date. Please run ./hack/update-crds.sh"
+  exit 1
+  fi
```
wallrj: Indentation looks off.

cassandra-operator/hack/verify-crds.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
wallrj: Maybe add a note that this is based off: https://github.com/jetstack/cert-manager/blob/master/hack/verify-crds.sh

cassandra-operator/hack/verify-crds.sh
```diff
@@ -0,0 +1,32 @@
...
+  rm -rf "${output}"
+}
+trap "cleanup" EXIT SIGINT
+
+tmp="${output}/cassandra-operator"
+
+
+rsync -avvL "${REPO_ROOT}"/ "${tmp}" >/dev/null
+cd "${tmp}"
+"./hack/update-crds.sh"
```
wallrj: In the original script, this is parameterized so that it can be used to verify any update-XXX.sh script.
Shall we do that too?

cassandra-operator/hack/verify-crds.sh
```diff
@@ -0,0 +1,32 @@
...
+
+echo "diffing against freshly generated crd (${tmp})"
```
wallrj: I'd have thought that you could use `if ! diff .... ` here, but if it works for cert-manager, fine.

cassandra-operator/Makefile
```diff
@@ -142,4 +142,12 @@ generate-types:
...
 	$(projectDir)/hack/update-codegen.sh
 
+generate-crds:
+	@echo "== generate-crds"
+	$(projectDir)/hack/update-crds.sh
+
+validate-crds:
+	@echo "== validate-crds"
+	$(projectDir)/hack/verify-crds.sh
+
```
wallrj: Strictly, these need to be added to the list of `.PHONY:` targets at the top of the makefile.

cassandra-operator/Makefile
```diff
@@ -142,4 +142,12 @@ generate-types:
 	@echo "== generate-types"
 	$(projectDir)/hack/update-codegen.sh
 
+generate-crds:
```
wallrj: Should we depend on another make target that installs all the go mod dependencies?

sebbonnet: We should also make `deploy` target depend on this, so that our e2e test will run against the freshly generated CRD

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
```diff
@@ -1,13 +1,538 @@
...
 spec:
   group: core.sky.uk
   names:
     kind: Cassandra
-    listKind: CassandraList
     plural: cassandras
-    singular: cassandra
-  scope: Namespaced
-  version: v1alpha1
+  scope: ""
```
sebbonnet: Cassandra is a namespaced resource, we should find a way to set the scope.

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
```diff
@@ -1,13 +1,538 @@
...
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: Cassandra defines a Cassandra cluster
```
sebbonnet: We should add description to all our fields, I'll raise a follow issue for that

sebbonnet:
I was going to add that we should update the README to mention how to configure the cluster (thinking `--feature-gates=CustomResourceValidation`) to benefit from this feature, but that's probably something for the WIKI and we shouldn't update it until we release this change. 

In anycase we should update the CHANGELOG.md too, as this is a new feature.

sebbonnet:
Actually don't worry about the CHANGELOG for now. Probably best if we deal with it when releasing a new version

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
kragniz: `  scope: ""` results in a `Namespaced` resource. It's a bit ugly, but I can't seem to coerce controller-gen into making it look nicer

cassandra-operator/hack/verify-crds.sh
kragniz: I'm not sure that adds much benefit, other than being easier to copy+paste to other scripts

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
sebbonnet: really, that's odd. I thought the default scope was cluster-scope, so are we saying that `scope: ""` is not the same as not setting the `scope` and actually translate to Namespaced?
Is there an issue on controller-gen to track this problem?

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
sebbonnet: I've raised #95 as a follow-up

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
kragniz: I've filed https://github.com/kubernetes-sigs/controller-tools/issues/214

kragniz:
Looks like the e2e tests are hitting this https://github.com/kubernetes/kubernetes/issues/59485

cassandra-operator/Makefile
```diff
@@ -46,7 +46,8 @@ setup:
 		github.com/sky-uk/licence-compliance-checker \
 		github.com/onsi/ginkgo/ginkgo \
 		golang.org/x/tools/cmd/goimports \
-		golang.org/x/lint/golint
+		golang.org/x/lint/golint \
+		sigs.k8s.io/controller-tools/cmd/controller-gen
```
sebbonnet: can we make `setup` depend on `ensure-go-modules`.
I got caught when testing this locally, so think that would be a useful thing to have

sebbonnet: we should also add that dependency on the `cassandra-snapshot` module too

kragniz:
@sebbonnet I've added a couple of tests

cassandra-operator/Makefile
kragniz: done

cassandra-operator/Makefile
```diff
@@ -31,7 +31,7 @@ install: build push-image
 
 check-style: fmt vet lint
 
-check: check-style validate-licences test deploy-operator integration-test e2e-test
+check: check-style validate-licences validate-crds test deploy-operator integration-test e2e-test
```
wallrj: I think  `validate-crds` ought to be part of the `check-style` target.

cassandra-operator/Makefile
```diff
@@ -119,7 +120,7 @@ docker-image: go-build common-docker-build
 
 push-image: docker-image common-docker-push
 
-deploy-operator:
+deploy-operator: generate-crds
```
wallrj: If `validate-crds` has been called earlier, by `check-style`, this should not be necessary.
In Kubernetes (and Cert Manager) we leave it up to the author to run the code generation scripts.
Otherwise, the CI may be running generated code which never gets checked into the repository.

cassandra-operator/test/e2e/parallel/validation/testdata/invalid-value-spec.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: invalidfieldcluster
+spec:
+  pod:
+    image: "cassandra:3.11"
+    cpu: 1m
+    memory: 1Gi
+    storageSize: 100Mi
+  racks:
+  - name: a
+    replicas: "two"
```
wallrj: maybe add a comment, highlighting that this is the invalid field.

cassandra-operator/Makefile
sebbonnet: Good point.
So we're saying that `generate-crds` and `generate-types` are targets that should be manually invoked? If so, then perhaps add a comment to say so

cassandra-operator/Makefile
sebbonnet: I'm not sure that makes sense. `check-sytle` to me is about validating the code conforms to coding guidelines, whereas `validate-crds` is validating that the CRD definition is in sync with the code.

sebbonnet: Were you suggesting that, so you get a fast feedback before running the 2e2 tests?
I think that all pre-e2e-test like`check-style validate-licences validate-crds test` should be moved to the build target  - that is something we can address in https://github.com/sky-uk/cassandra-operator/pull/99

PR #99, opened 2019-05-28
Run all static checks and compilation steps before launching the e2e tests

wallrj:
Closes: #66 
Supplants: #83


.gitignore
```diff
@@ -10,3 +10,4 @@ build
 out
 bin/
 .project
+.envrc
```
sebbonnet: newline missing

Makefile
```diff
@@ -15,12 +15,16 @@ GINKGO_COMPILERS ?= 0
 gitRev := $(shell git rev-parse --short HEAD)
 projectDir := $(realpath $(dir $(firstword $(MAKEFILE_LIST))))
 
-travis: make-all
+all: setup check-style build dind install check
```
sebbonnet: I would change this to: `all: setup build install dind check`, as:
- dind is only required before running the tests
- I would integrate the `check-style` target in the `build`

cassandra-operator/Makefile
```diff
@@ -23,7 +23,7 @@ testPkgs := $(shell go list ./... | grep /test/e2e)
 
 include $(projectDir)/../MakefileDocker
 
-travis: clean install check
+all: clean install check
 
 build: go-build go-install docker-image
```
sebbonnet: `go-install` should move to the install target: `install: build go-install push-image`

sebbonnet: I would have build like that: `build: go-build check-style docker-image`
I would mean check-style are run before building the docker image. 

We could even extend it to run the licence compliance checks here too and unit tests.
So perhaps this would be even better as `build: go-build check-style validate-licences test docker-image`.
This would add a little bit of duplication with the `check` target, but I think this would provide the fastest feedback, before moving onto the more expensive `deploy` and `e2e-test` targets.
What do you think?

cassandra-snapshot/Makefile
```diff
@@ -20,7 +20,7 @@ testPkgs := $(shell go list ./... | grep /test/e2e)
 
 include $(projectDir)/../MakefileDocker
 
-travis: clean install check
+all: clean install check
 
 build: go-build docker-image
```
sebbonnet: same comment as per cassandra-operator

Makefile
```diff
@@ -73,7 +84,7 @@ setup-all:
 	$(MAKE) -C cassandra-snapshot setup
 	$(MAKE) -C cassandra-operator setup
 
-recreate-dind-cluster:
+dind:
 	@echo "== recreate dind cluster"
 	NAMESPACE=$(NAMESPACE) $(projectDir)/test-kubernetes-cluster/recreate-dind-cluster.sh
```
sebbonnet: not sure if `check-style-all` is useful anymore if `check-style` is invoked as part of the `build` target?

wallrj: I find it useful to be able to run make check-style for a quick sanity check.

Makefile
wallrj: I've check-style, validate-licences , validate-crds to the build target.

cassandra-operator/Makefile
wallrj: I've moved the style and validation checks from `check` to `build`.
`check` now only runs tests (which are pre-compiled as part of the build) .

Makefile
wallrj: See above

wallrj:
I've started the Docker registry early, during `setup` so that we can simply call `make install` before launching the dind cluster and running the tests.

Makefile
```diff
@@ -15,9 +15,10 @@ GINKGO_COMPILERS ?= 0
 gitRev := $(shell git rev-parse --short HEAD)
 projectDir := $(realpath $(dir $(firstword $(MAKEFILE_LIST))))
 
-all: setup check-style build dind install check
+all: setup install dind check
 
 setup: check-system-dependencies setup-all
+	docker run -d --name=dind-registry --rm -p 5000:5000 registry:2
```
sebbonnet: Would this not complain that the registry is already running when starting the dind cluster?

sebbonnet: ah no wait, `recreate-dind-cluster.sh` would not attempt to start the registry if it finds it already running

sebbonnet: picky comment: would be nicer to move to docker run to a new `run-local-registry` target

cassandra-bootstrapper/Makefile
```diff
@@ -9,11 +9,11 @@ include $(projectDir)/../MakefileDocker
 
 all: clean install check
 
-build: jar docker-image
+build: check-style jar docker-image
```
sebbonnet: I suppose `check-sytle` would eventually be integrated in the build.gradle `jar` task. But leaving it here for now doesn't hurt

cassandra-operator/Makefile
```diff
@@ -25,13 +25,13 @@ include $(projectDir)/../MakefileDocker
 
 all: clean install check
 
-build: go-build docker-image
+build: check-style validate-licences validate-crds go-build docker-image
```
sebbonnet: Shouldn't `go-build` be called before `validate-licences` - this is to avoid missing a dependency that would be added during the go build
Shouldn't `go-build` be called before `validate-crds`?

cassandra-snapshot/Makefile
```diff
@@ -22,13 +22,13 @@ include $(projectDir)/../MakefileDocker
 
 all: clean install check
 
-build: go-build docker-image
+build: check-style validate-licences go-build docker-image
```
sebbonnet: Same question as above: Shouldn't go-build be called before validate-licences - this is to avoid missing a dependency that would be added during the go build

PR #101, opened 2019-05-30
Release on tag or merge to master

sebbonnet:
- On merge to master, push a release-candidate
Run a full build and push docker images labelled with the current git
revision to the release repository.
This is to make it possible to run additional tests in an integrated
environment to validate whether the release-candidate is ready to release.

- On tag, promote the release-candidate to release
Skip the build, but promote the release-candidate for the current git
revision to release.
Use the tag to label the docker images and push them to the release
repository. Also updates the latest images with this version.


sebbonnet:
This needs reviewing following changes brought in https://github.com/sky-uk/cassandra-operator/pull/102

sebbonnet:
I've reviewed this PR to conditionally add stages:
- `test` stage is not run on tag
- `deploy` stage is only run on push to master - on merge to master and tags

MakefileDocker
```diff
@@ -28,12 +29,16 @@ common-docker-clean:
 common-docker-release:
 	@echo "== common-docker-release"
 ifeq ($(strip $(gitTag)),)
-	@echo "no tag on $(gitRev), skipping release"
+	@echo "Releasing candidate based on rev: $(gitRev). Docker image: $(dockerReleaseCandidateImage)"
```
alangibson01: All this seems very common - is there a case for splitting it out to a script or something like that which is more parameterised?

sebbonnet: Do you mean the image promotion? I could extract that to a script, but not sure how to determine the path to the script as this makefile is "included" and so the current path would be that of the sub-module.

sebbonnet:
@alangibson01 I've extracted the image promotion to a script, but I'm not sure it made things simpler
Given that MakefileDocker is included in other Makefile, I had to make use of the `lastword` and also wrap the script call in a function to avoid the output printing to the console

PR #107, opened 2019-06-03
Split up release and release-candidate stages

sebbonnet:
As we need to upload local artifacts on merge to master, but only
require promote existing images on tag

Also change `run-local-registry` to run when needed

Relates to #101


sebbonnet:
Example of release promotion: https://travis-ci.com/sky-uk/cassandra-operator/builds/114957760

PR #108, opened 2019-06-04
Skip travis script phase

sebbonnet:
Which defaults to running `make` for go language

This is not an issue for jobs that already have a `script` defined, as
their definition will overwrite the default one.
However as the deploy jobs do not have a script defined (they have a
deploy struct using a script - that is not the same thing), the travis
default script is invoked which is not what we want.

So overwrite the `default` script at the top level rather than for each
deploy job.

Follows on from #107


sebbonnet:
See https://travis-ci.com/sky-uk/cassandra-operator/jobs/205218192 for an example of the script phase being invoked

sebbonnet:
Example of release promotion: https://travis-ci.com/sky-uk/cassandra-operator/builds/114957760

PR #110, opened 2019-06-04
Use Kubectl wrapper function rather than command

sebbonnet:
So the correct kube context, config and namespace is used when running
`kubectl` commands. The wrapper command honours the various env variables that can be used to customise context and namespace where the tests are run.


sebbonnet:
Not sure, I'll have a look in a bit

Issue #104, opened 2019-05-31
Verify that fake Cassandra responds the same as real Cassandra


wallrj:
See: https://pythonspeed.com/articles/verified-fakes/

If we're going to use a Fake / Stub Cassandra, we ought to have tests that ensure that it behaves the same as the real thing.

We could add a test suite that is run against both the real and the fake Cassandra.

We'd need to test the fake Jolokia behaviour and the fake nodetool behaviour.

See #32

wallrj:
There are some existing projects that we might be able to use / re-use or learn from:
 * https://github.com/datastax/simulacron
 * https://www.scassandra.org/
 * https://www.testcontainers.org/modules/databases/cassandra/

adamdougal:
Closing as we only use fake cassandra where it doesn't matter how Cassandra responds.

PR #111, opened 2019-06-05
Add tests which verify the fake-cassandra-docker image

wallrj:
Part of: #104 

This implements some integration / e2e style tests for fake cassandra docker image.
I point the existing metrics client at the fake and ensure that it gives a sufficient response to get the status of a node.
A follow up would be to also perform this test against a real cassandra docker image (with jolokia installed).

I'd quite like to thin out the goss tests to only verify that the known faked endpoints respond with HTTP 200 codes.

If this is accepted and merged, I'll rebase #10 on top, and extend the fake so that it can respond to liveness and readiness checks from the sidecar.


fake-cassandra-docker/test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,104 @@
...
+		session.Interrupt()
+		Eventually(session, "5s").Should(Exit())
+	})
+
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "e2e", test.CreateSequentialReporters("e2e"))
+}
+
+var _ = Context("Cassandra Docker container", func() {
+	It("should accept Jolokia requests", func() {
```
sebbonnet: we should rephrase this to express the requirement a bit better - perhaps add a `Context("Jolokia")` with a `It("should respond with a node up and normal")` or similar

fake-cassandra-docker/goss.yaml
```diff
@@ -21,15 +21,27 @@ http:
...
-    body: ["status\":200"]
+    body: ['status":200']
   "http://localhost:7777/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getDatacenter/localhost":
     status: 200
     body: ["status\":403"]
-  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes,UnreachableNodes,JoiningNodes,LeavingNodes,MovingNodes":
+  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes":
     status: 200
-    body: ["status\":200"]
+    body: ['status":200']
```
sebbonnet: Shouldn't this also return a value with the node ip address?

fake-cassandra-docker/Makefile
```diff
@@ -31,10 +44,27 @@ push-image: docker-image common-docker-push
...
 
 clean: common-docker-clean
 	@echo "== clean"
 	$(gradle) :fake-cassandra-docker:clean
 
 release: common-docker-release
+
+e2e-test: ensure-test-report-dir-exists
+	@echo "== e2e-test"
+	KUBE_CONTEXT=$(KUBE_CONTEXT) \
```
sebbonnet: I don't think any of these environment variables are used.

fake-cassandra-docker/Makefile
```diff
@@ -31,10 +44,27 @@ push-image: docker-image common-docker-push
...
+			NAMESPACE=$(NAMESPACE) \
+			POD_START_TIMEOUT=$(POD_START_TIMEOUT) \
+			FAKE_CASSANDRA_IMAGE=$(FAKE_CASSANDRA_IMAGE) \
+			USE_MOCK=$(USE_MOCK) \
+			CASSANDRA_SNAPSHOT_IMAGE=$(CASSANDRA_SNAPSHOT_IMAGE) \
+			CASSANDRA_BOOTSTRAPPER_IMAGE=$(CASSANDRA_BOOTSTRAPPER_IMAGE) \
+			ginkgo -p -nodes=$(GINKGO_NODES) -r -compilers=$(GINKGO_COMPILERS) --v --progress -stream -timeout=30m test/e2e -- -junit-report-dir $(junitReportDir)
```
sebbonnet: Your should be passing the `NODE_LISTEN_ADDRESS=192.0.2.100` variable here too, so the tests can use it

fake-cassandra-docker/test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func init() {
+	fakeCassandraImage = os.Getenv("FAKE_CASSANDRA_IMAGE")
+	if fakeCassandraImage == "" {
+		panic("FAKE_CASSANDRA_IMAGE must be supplied")
+	}
+}
+
+var nodeListenAddress = "192.0.2.100"
```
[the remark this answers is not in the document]
sebbonnet: forget it, i just noticed the value doesn't matter as it is passed to the docker run

fake-cassandra-docker/Makefile
sebbonnet: forget it, i just noticed the value doesn't matter as it is passed to the docker run in the test

wallrj:
@sebbonnet I've implemented a much simpler change, without the additional tests in https://github.com/sky-uk/cassandra-operator/pull/115, as you suggested above and as discussed at the standup this morning.

PR #115, opened 2019-06-07
Fake Cassandra Jolokia responses

wallrj:
Required for: #98 

This makes the fake-cassandra-docker container respond with more realistic Jolokia responses.
Enough to make the metrics client code think that the local node is UP and NORMAL.

### DIFF
--- a/fake-cassandra-docker/goss.yaml
+++ b/fake-cassandra-docker/goss.yaml
@@ -21,15 +21,29 @@ http:
   "http://localhost:7777/jolokia/read/java.lang:type=Memory/HeapMemoryUsage":
     status: 200
     body: ["status\":403"]
-  "http://localhost:7777/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getRack/localhost":
+  "http://localhost:7777/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getRack/192.0.2.100":
     status: 200
-    body: ["status\":200"]
+    body: ['status":200']
   "http://localhost:7777/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getDatacenter/localhost":
     status: 200
     body: ["status\":403"]
-  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes,UnreachableNodes,JoiningNodes,LeavingNodes,MovingNodes":
+  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes":
     status: 200
-    body: ["status\":200"]
+    body:
+      - '"status":200'
+      - '"value":["192.0.2.100"]'
+  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/UnreachableNodes":
+    status: 200
+    body: ['status":200']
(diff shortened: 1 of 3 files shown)


fake-cassandra-docker/goss.yaml
```diff
@@ -21,15 +21,29 @@ http:
...
+    body: ['status":200']
   "http://localhost:7777/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getDatacenter/localhost":
     status: 200
     body: ["status\":403"]
-  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes,UnreachableNodes,JoiningNodes,LeavingNodes,MovingNodes":
+  "http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes":
     status: 200
-    body: ["status\":200"]
+    body:
+      - '"status":200'
```
sebbonnet: So body can also use patterns, nice!

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -9,16 +9,22 @@
...
-import java.util.Set;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 public class FakeCassandra {
     private static final int[] DUMMY_PORTS = new int[]{7000, 7199, 9042};
+    public static String nodeListenAddress;
```
sebbonnet: This should have been private

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -106,12 +112,22 @@ public Response serve(final IHTTPSession session) {
...
-            "/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes,UnreachableNodes,JoiningNodes,LeavingNodes,MovingNodes"
-    )));
-
+    private static Map<String, String> PERMITTED_PATHS;
+    static {
+        PERMITTED_PATHS = new HashMap<>();
+        PERMITTED_PATHS.put("/jolokia/exec/org.apache.cassandra.db:type=EndpointSnitchInfo/getRack/" + FakeCassandra.nodeListenAddress,
+                            "{\"status\":200}");
+        PERMITTED_PATHS.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes",
+                            "{\"status\":200, \"value\":[\"" + FakeCassandra.nodeListenAddress+ "\"]}");
```
sebbonnet: Technically the response body contains more data, but in our case we are only interested in status and value.
For the record here is the full payload when running `curl http://localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes | jq`
```
{
  "request": {
    "mbean": "org.apache.cassandra.db:type=StorageService",
    "attribute": "LiveNodes",
    "type": "read"
  },
  "value": [
    "172.20.2.8"
  ],
  "timestamp": 1559923005,
  "status": 200
}

```

Issue #30, opened 2018-10-16
Pod liveness and readiness probes frequently start JVMs, consuming CPU


alangibson01:
Cassandra pod liveness and readiness probes both use `nodetool` in order to perform their checks. This is rather heavyweight, as it requires starting up a JVM. Given we run 2 liveness checks and 4 readiness checks a minute, this demands rather a lot of CPU. Looking at the CPU metrics for the cluster belonging to verification-app in dev, we can see that CPU usage hovers around 50% even when the cluster is at rest:

![screenshot from 2018-10-16 15-01-26](https://user-images.githubusercontent.com/20461485/47022452-6dd68200-d155-11e8-8354-9c5fbd1c8ef8.png)

wallrj:
Here's how we tackled this in Navigator: https://github.com/jetstack/navigator/pull/170

sebbonnet:
Nice to see you cracked this problem before :)
I think we both agree that using jolokia as a nodetool replacement sounds like the right approach.

As we have jolokia installed on cassandra nodes already, I imagined initially that we could simply use a curl command to find a node status. However as the information to determine the node state (e.g. normal) is scattered across different mbean attributes I think we will need write a small tool to combine the results from multiple mbean responses.

To illustrate my point:
- node UP: 
```
curl localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes | grep <pod-ip>
```
- node normal: 
```
!curl localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/JoiningNodes | grep -v <pod-ip> && \
!curl localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/LeavingNodes | grep -v <pod-ip> && \
!curl localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/MovingNodes | grep -v <pod-ip> && \
!curl localhost:7777/jolokia/read/org.apache.cassandra.db:type=StorageService/UnreachableNodes | grep -v <pod-ip> &&
```

We already use jolokia to expose cluster level metrics in the operator (https://github.com/sky-uk/cassandra-operator/blob/master/cassandra-operator/pkg/metrics/metrics.go#L85), so it would be nice it we could extract that logic and use it in the "probe" tool. We'll have to see whether it's worth the effort.

So i think the plan of action for me would be:
- extract the jolokia nodestatus logic to its own module
[...]

PR #116, opened 2019-06-07
Cassandra sidecar which responds to liveness and readiness probes

wallrj:
Closes: #30 
Supplants: #98 
Stacked upon: #115 

Co-authored-by: @kragniz


cassandra-operator/pkg/metrics/suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/test"
+)
+
+func TestMetrics(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "Metrics Suite", test.CreateParallelReporters("metrics"))
```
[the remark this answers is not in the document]
sebbonnet: ah wait, it is used by both `ready_test` and `metrics_test`.
However that means that the suite is also being used by `nodestatus_test` when it shouldn't. 
It's not ideal, but i think it simplifies things, so that's ok

cassandra-operator/test/e2e/matchers.go
```diff
@@ -153,14 +155,30 @@ func (matcher *haveEvent) NegatedFailureMessage(actual interface{}) (message str
...
+		containers: func(p coreV1.Pod) []coreV1.Container {
+			return p.Spec.Containers
+		},
+	}
+}
+
+//
+// HaveInitContainer matcher
+//
+func HaveInitContainer(expected interface{}) types.GomegaMatcher {
```
sebbonnet: I don't think this is used anywhere

cassandra-operator/test/e2e/parallel/creation/creation_test.go
```diff
@@ -100,8 +101,7 @@ var _ = Context("When a cluster with a given name doesn't already exist", func()
 
 		By("creating pods with the specified resources")
 		Expect(PodsForCluster(Namespace, multipleRacksCluster.Name)()).Should(Each(And(
-			HaveASingleContainer(ContainerExpectation{
-				BootstrapperImageName:          CassandraBootstrapperImageName,
+			HaveContainer(ContainerExpectation{
```
sebbonnet: I think we should also check that we have a "sidecar" container

cassandra-operator/test/e2e/test_environment.go
```diff
@@ -2,17 +2,18 @@ package e2e
...
 	"fmt"
-	log "github.com/sirupsen/logrus"
-	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1"
 	"os"
 	"os/exec"
 	"strings"
 	"time"
 
-	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/client/clientset/versioned"
+	log "github.com/sirupsen/logrus"
```
sebbonnet: good catch

cassandra-sidecar/cmd/main.go
```diff
@@ -0,0 +1,130 @@
...
+		nil,
+	)
+
+	http.HandleFunc("/ready", func(w http.ResponseWriter, r *http.Request) {
+		logger := logger.WithFields(
+			log.Fields{
+				"endpoint": "/ready",
+			},
+		)
+		logger.Debug("request received")
```
sebbonnet: Is that necessary? Maybe add the name of the endpoint, so we can differenciate from the liveness check

cassandra-sidecar/cmd/main.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	})
+
+	http.HandleFunc("/live", func(w http.ResponseWriter, r *http.Request) {
+		logger := logger.WithFields(
+			log.Fields{
+				"endpoint": "/live",
+			},
+		)
+		logger.Debug("request received")
```
sebbonnet: Maybe add the name of the endpoint, so we can differenciate from readiness check

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -9,16 +9,22 @@
 import java.net.Socket;
 import java.util.Arrays;
 import java.util.Collections;
```
sebbonnet: Collections is no longer used, can we remove that import

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -55,6 +62,20 @@ var defaultReadinessProbe = v1alpha1.Probe{
...
+	sidecarMemoryLimit   resource.Quantity
+	sidecarCPURequest    resource.Quantity
+	sidecarCPULimit      resource.Quantity
+)
+
+func init() {
+	sidecarMemoryRequest = resource.MustParse(defaultSidecarMemoryRequest)
+	sidecarMemoryLimit = resource.MustParse(defaultSidecarMemoryLimit)
+	sidecarCPURequest = resource.MustParse(defaultSidecarCPURequest)
+	sidecarCPULimit = resource.MustParse(defaultSidecarCPULimit)
```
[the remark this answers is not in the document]
wallrj: I've made it so that `Cassandra.Spec.Pod.CPU` is used, if it is lower than than `100m`.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -515,20 +549,45 @@ func (c *Cluster) createCassandraContainer(rack *v1alpha1.Rack, customConfigMap
 			},
 		},
 		Resources:      c.createResourceRequirements(),
-		LivenessProbe:  createProbe(c.definition.Spec.Pod.LivenessProbe, "/bin/sh", "-c", "nodetool info"),
-		ReadinessProbe: createProbe(c.definition.Spec.Pod.ReadinessProbe, "/bin/sh", "-c", "nodetool status | grep -E \"^UN\\s+${NODE_LISTEN_ADDRESS}\""),
+		LivenessProbe:  createHTTPProbe(c.definition.Spec.Pod.LivenessProbe, "/live", 8080),
```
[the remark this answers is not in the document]
wallrj: Done. I've added a new command line argument `--health-server-port` which is set from the same constant used in the probe definitions.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -515,20 +549,45 @@ func (c *Cluster) createCassandraContainer(rack *v1alpha1.Rack, customConfigMap
...
+		ReadinessProbe: createHTTPProbe(c.definition.Spec.Pod.ReadinessProbe, "/ready", 8080),
 		Lifecycle: &v1.Lifecycle{
 			PreStop: &v1.Handler{
 				Exec: &v1.ExecAction{
 					Command: []string{"/bin/sh", "-c", "nodetool drain"},
 				},
 			},
 		},
-		Env:          []v1.EnvVar{{Name: "EXTRA_CLASSPATH", Value: "/extra-lib/cassandra-seed-provider.jar"}},
+		Env:          env,
```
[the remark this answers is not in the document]
wallrj: Cassandra doesn't, but `fake-cassandra-docker` had no way of knowing what node address it was faking. This seemed like the easiest way.
I've changed it now to read the `listen_address` from cassandra.yaml.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -515,20 +549,45 @@ func (c *Cluster) createCassandraContainer(rack *v1alpha1.Rack, customConfigMap
...
 		VolumeMounts: c.createVolumeMounts(customConfigMap),
 	}
 }
 
+func (c *Cluster) createCassandraSidecarContainer(rack *v1alpha1.Rack) v1.Container {
+	return v1.Container{
```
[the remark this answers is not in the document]
wallrj: I'm not convinced of the value of this test, but ok. I've added a new Ginkgo test for the three env vars required by the sidecar.

cassandra-operator/pkg/metrics/ready.go
```diff
@@ -0,0 +1,56 @@
...
+	gatherer := NewGatherer(n.urlProvider, &Config{
+		RequestTimeout: 20 * time.Second,
+	})
+	clusterStatus, err := gatherer.GatherMetricsFor(n.cluster)
+	if err != nil {
+		return false, err
+	}
+	nodeStatus := clusterStatus.NodeStatus(host)
+	if nodeStatus == nil {
+		return false, fmt.Errorf("node not found: %s", host)
```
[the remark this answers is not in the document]
wallrj: Good spot. Now added.

cassandra-operator/test/e2e/parallel/creation/creation_test.go
wallrj: Yeah, I considered it, but it seems pointless here.
We've already got unit tests that demonstrate that a sidecar container is added to the statefulset template.
The sidecar container is an implementation detail of how we implement liveness / readiness probes.
The creation and scale-out tests already demonstrate that the readiness probes are working properly. 
An e2e test for liveness probes would be a good addition, but I think that can be deferred for another PR.

It would also require refactoring the `ContainerExpectation` code, because that expects every container to have a defined liveness / readiness timeouts etc.

cassandra-sidecar/cmd/main.go
wallrj: Yeah, it was helpful while debugging this branch.
The name of endpoint is already present as a logrus field, above.

cassandra-sidecar/cmd/main.go
wallrj: It's already added as a logrus field, above.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
wallrj: Done.

wallrj: Also added some better logging and changed it so that it reads nodeListenAddress from cassandra.yaml, which avoids having to supply NODE_LISTEN_ADDRESS as a pod env var.

cassandra-operator/test/e2e/matchers.go
wallrj: The original matcher (`HaveASingleContainer`) checked both the main container, and certain fields of  the init containers.
I was originally going to preserve that test for the init-containers, but it'd require some refactoring of the `ContainerExpectation` code.
I don't think it's useful to test the fields of init containers in the e2e tests since we already test that in the unit tests for `createStatefulSetForRack`.
There are already  E2E tests  which check that configmaps can be supplied and that they result in altered Cassandra config.
The bootstrap init containers are just an implementation detail.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -17,19 +21,33 @@
...
 public class FakeCassandra {
     private static final int[] DUMMY_PORTS = new int[]{7000, 7199, 9042};
+    private static final String CASSANDRA_YAML_PATH = "/etc/cassandra/cassandra.yaml";
+    private static final Logger logger = LoggerFactory.getLogger(FakeCassandra.class);
+
     public static String nodeListenAddress;
 
     public static void main(String[] args) {
-        nodeListenAddress = System.getenv("NODE_LISTEN_ADDRESS");
+        InputStream input;
```
sebbonnet: This needs tidying up:
- we should extract the node address lookup to a method
- add the node addresss to the FakeCassandra constructor and pass it down to the FakeJolokiaServer that would also accept it as a constructor param
- PERMITTED_PATHS would also be initialised in the  FakeJolokiaServer constructor based on the node address

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: I like this approach better :+1:

cassandra-operator/test/e2e/parallel/creation/creation_test.go
sebbonnet: Fair enough

cassandra-operator/test/e2e/matchers.go
sebbonnet: Agreed - checking init containers is best done as unit test level

cassandra-operator/pkg/cluster/cluster.go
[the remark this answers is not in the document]
sebbonnet: we need the same logic for memory, as it could be set to 0 too (even though i would not recommend it)

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -577,12 +573,15 @@ func (c *Cluster) createCassandraSidecarContainer(rack *v1alpha1.Rack) v1.Contai
 		Env: c.createEnvironmentVariableDefinition(rack),
 		Resources: v1.ResourceRequirements{
 			Requests: v1.ResourceList{
-				v1.ResourceCPU:    sidecarCPURequest,
+				v1.ResourceCPU: minQuantity(
+					c.definition.Spec.Pod.CPU,
+					maxSidecarCPURequest,
+				),
 				v1.ResourceMemory: sidecarMemoryRequest,
```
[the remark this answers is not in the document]
wallrj: Done. But worth noting that the `cluster.validatePodResources` currently prevents `Cassandra.Spec.Pod.Memory: 0`

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
wallrj: Done. Quite interesting to get to know Java better.

sebbonnet:
Tidied up fake cassandra a bit :) https://github.com/sky-uk/cassandra-operator/pull/121

Issue #46, opened 2019-03-28
Define style guide for java modules


sebbonnet:
### Is your feature request related to a problem? Please describe
There are no style guide defined for the java modules meaning that enforcing code standard happens too late once the PR is submitted and requires a human to review the style which is both tedious and inconsistent.  

 ### Describe the solution you'd like
Add checkstyle as part of the build for both our java modules.
Perhaps try to harvest bits we like from https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/checkstyle.xml

wallrj:
This could be executed by `make -C cassandra-bootstrapper verify` which I added in #68

PR #121, opened 2019-06-12
Tidy up fake cassandra

sebbonnet:
To get rid off the compilation warnings


fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
-        PERMITTED_PATHS.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/UnreachableNodes",
-                            "{\"status\":200}");
-        PERMITTED_PATHS.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/JoiningNodes",
-                            "{\"status\":200}");
-        PERMITTED_PATHS.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/LeavingNodes",
-                            "{\"status\":200}");
-        PERMITTED_PATHS.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/MovingNodes",
-                            "{\"status\":200}");
-     }
+    final Map<String, String> permittedPaths;
```
wallrj: Aha, I didn't understand how  `final` worked. So it allows the variable to be assigned to *once* only.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
+
+    FakeJolokiaServerConfig(String nodeListenAddress) {
```
wallrj: Elsewhere, you've declared the constructor to be `private`. What's the rule of thumb?

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
+        permittedPaths.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/MovingNodes",
+                "{\"status\":200}");
+    }
 
+    @Override
     public String toString() {
-        return String.format("PERMITTED_PATHS: %s", PERMITTED_PATHS);
+        return "FakeJolokiaServerConfig{" +
+                "permittedPaths=" + permittedPaths +
+                '}';
```
wallrj: Much better.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
 }
 
 class FakeCassandraConfig {
-    private static final Logger logger = LoggerFactory.getLogger(FakeCassandraConfig.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(FakeCassandraConfig.class);
+    private static final String CASSANDRA_YAML_PATH = "/etc/cassandra/cassandra.yaml";
+    private static final String DEFAULT_NODE_ADDRESS = "localhost";
+
+    final int[] dummyPorts = new int[]{7000, 7199, 9042};
+    final String nodeListenAddress;
```
wallrj: Should all `final` variables be in UPPER_CASE?

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
+                "{\"status\":200}");
+        permittedPaths.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/JoiningNodes",
+                "{\"status\":200}");
+        permittedPaths.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/LeavingNodes",
+                "{\"status\":200}");
+        permittedPaths.put("/jolokia/read/org.apache.cassandra.db:type=StorageService/MovingNodes",
+                "{\"status\":200}");
+    }
 
+    @Override
```
wallrj: I need to read about what this does. It seemed to work without it.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: rule of thumb is to make the scope as restrictive as possible, so private, (default), protected, public. In this case the class is instantiates from the same package, so default scope is required

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: auto-generated by IntelliJ IDEA :)

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: no, only constants - that would be `static final` whatever their scope

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: this is to make sure the superclass method is actually overridden - it is doesn't it won't compile

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
sebbonnet: yes - for primitives the value cannot change, for object the reference (pointer) cannot change once assigned

sebbonnet:
If we had had a code-style in place, a lot of these would have been flagged during the build - see #46

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
-                             CASSANDRA_YAML_PATH,
-                             NODE_LISTEN_ADDRESS);
+        return "FakeCassandraConfig{" +
+                "dummyPorts=" + Arrays.toString(dummyPorts) +
+                ", nodeListenAddress='" + nodeListenAddress + '\'' +
+                '}';
     }
 }
 
 public class FakeCassandra {
```
alangibson01: In keeping with idiomatic Java, let's move the public class to be the first one declared in the file.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
+    private static final String CASSANDRA_YAML_PATH = "/etc/cassandra/cassandra.yaml";
+    private static final String DEFAULT_NODE_ADDRESS = "localhost";
+
+    final int[] dummyPorts = new int[]{7000, 7199, 9042};
+    final String nodeListenAddress;
 
-    static final int[] DUMMY_PORTS = new int[]{7000, 7199, 9042};
-    static final String CASSANDRA_YAML_PATH = "/etc/cassandra/cassandra.yaml";
-    String NODE_LISTEN_ADDRESS = "localhost";
+    static FakeCassandraConfig load() throws FileNotFoundException {
```
alangibson01: I'd prefer it if this was a separate no-arg constructor, rather than a static method.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
 
     public static void main(String[] args) {
         FakeCassandraConfig cassandraConfig;
         try {
-            cassandraConfig = FakeCassandraConfig.Load();
+            cassandraConfig = FakeCassandraConfig.load();
         } catch (FileNotFoundException e) {
-            logger.error("failed to load configuration", e);
+            LOGGER.error("failed to load configuration", e);
             return;
```
alangibson01: Use `System.exit()` here with a non-zero exit code.

fake-cassandra-docker/src/main/java/com/sky/core/operators/fake/FakeCassandra.java
```diff
@@ -19,127 +19,129 @@
...
+        final FakeCassandra fc = new FakeCassandra(
+                cassandraConfig,
+                new FakeJolokiaServer(new FakeJolokiaServerConfig(cassandraConfig.nodeListenAddress)),
+                new FakeMetricsServer()
+        );
         try {
             fc.run();
         } catch (IOException e) {
-            logger.error("FakeCassandra startup failed", e);
+            LOGGER.error("FakeCassandra startup failed", e);
```
alangibson01: As above - use `System.exit()`.

Issue #72, opened 2019-05-08
Make all optional fields pointer types and +omitempty


kragniz:
Part of #33.

sebbonnet:
There is one struct left to do before we can close this:
```
type Probe struct {
	// +optional
	FailureThreshold int32 `json:"failureThreshold"`
	// +optional
	InitialDelaySeconds int32 `json:"initialDelaySeconds"`
	// +optional
	PeriodSeconds int32 `json:"periodSeconds"`
	// +optional
	SuccessThreshold int32 `json:"successThreshold"`
	// +optional
	TimeoutSeconds int32 `json:"timeoutSeconds"`
}
```

sebbonnet:
I don't think it makes sense to have `RetentionPolicy.Enabled` required. 
I think it should be optional and default to true, currently it defaults to false which means user would need to remember to set it before they can use the feature.
```
type RetentionPolicy struct {
	Enabled bool `json:"enabled"`
	// +optional
	RetentionPeriodDays *int32 `json:"retentionPeriodDays,omitempty"`
	// CleanupSchedule follows the cron format, see https://en.wikipedia.org/wiki/Cron
	CleanupSchedule string `json:"cleanupSchedule"`
	// +optional
	CleanupTimeoutSeconds *int32 `json:"cleanupTimeoutSeconds,omitempty"`
}
```

sebbonnet:
All optional fields have been dealt with

PR #126, opened 2019-06-14
Make Snapshot.RetentionPolicy.Enabled an optional field

wallrj:
Part of: #72


cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -69,7 +69,7 @@ func GetDatacenter(c *v1alpha1.Cassandra) string {
 
 // HasRetentionPolicyEnabled returns true when a retention policy exists and is enabled
 func HasRetentionPolicyEnabled(snapshot *v1alpha1.Snapshot) bool {
-	return snapshot.RetentionPolicy != nil && snapshot.RetentionPolicy.Enabled
+	return snapshot.RetentionPolicy != nil && *snapshot.RetentionPolicy.Enabled
```
wallrj: I could check for `snapshot.RetentionPolicy.Enabled == nil` here too, but I'd like to think that these functions are only ever used on clusters where defaulting has been applied.

sebbonnet: Let's be safe and add that check - it's simple enough that there's hardly any reason not to do it

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -98,6 +98,15 @@ func NewWithoutValidation(clusterDefinition *v1alpha1.Cassandra) *Cluster {
 	return cluster
 }
 
+func setDefaultSnapshotRetentionPolicy(clusterDefinition *v1alpha1.Cassandra) {
+	switch {
+	case clusterDefinition.Spec.Snapshot == nil:
```
sebbonnet: this may have been simpler to move the defaulting inside the `validateSnapshot` method to avoid the extra nil checks https://github.com/sky-uk/cassandra-operator/pull/126/files#diff-d3a32842a06dfdbc2a15ac4f7f73fdf9R217

sebbonnet: but on the other hand keeping apart the defaulting and validation apart is cleaner

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -48,14 +48,34 @@ var _ = Describe("cluster construction", func() {
...
 			},
 		}
 	})
 
+	Context("config defaulting", func() {
+		It("should default Cassandra.Spec.Snapshot.RetentionPolicy.Enabled to true", func() {
+			clusterDef.Spec.Snapshot.RetentionPolicy.Enabled = nil
+			cluster, err := ACluster(clusterDef)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cluster.definition.Spec.Snapshot.RetentionPolicy.Enabled).To(Not(BeNil()))
```
sebbonnet: this check is not necessary

cassandra-operator/test/e2e/parallel/snapshot/snapshot_test.go
```diff
@@ -74,7 +77,7 @@ var _ = Describe("Cassandra snapshot scheduling", func() {
 					Schedule:  "59 23 * * *",
 					Keyspaces: []string{"keyspace1", "keyspace3"},
 					RetentionPolicy: &v1alpha1.RetentionPolicy{
-						Enabled:               true,
+						Enabled:               ptr.Bool(true),
```
sebbonnet: I think we should remove the explicit "Enabled: true" for test cases as this is the default value.

cassandra-operator/pkg/cluster/cluster.go
wallrj: I considered moving it into the helpers package, but decided to leave that for another time.

wallrj:
Removing the `RetentionPolicy.Enabled` from the E2E test manifests revealed that the defaulting wasn't happening early enough.
The `cluster.New` (which previously performed the defaulting) was being called only when the `AddSnapshotCleanup` action got executed.

I've now updated the code so that the defaulting is performed immediately after the Cassandra objects are received from the informer callback functions.

Issue #34, opened 2019-03-22
Cluster config modification events are not generated 


sebbonnet:
### Describe the bug

Cluster config modification events are sometimes not appearing as Kubernetes events even though the operator code clearly is executed. This happens infrequently when running tests against either a DIND cluster or dev, but never when the failing tests are run in isolation.

This perhaps suggests that some events get overwritten or dropped. This is surprising given that the `k8s.io/client-go/tools/record/EventRecorder` can queue up to 1000 events and logs an error when the same even has reached the max retry count.

Failing test: Allowable cluster modifications > cluster config file changes
```
[10] Allowable cluster modifications
[10] /home/sebbonnet/gowork/src/github.com/sky-uk/core-operators/cassandra-operator/test/e2e/parallel/modification/modification_test.go:50
[10]   cluster config file changes
[10]   /home/sebbonnet/gowork/src/github.com/sky-uk/core-operators/cassandra-operator/test/e2e/parallel/modification/modification_test.go:175
[10]     should trigger a rolling restarts of the cluster stateful set when the custom config file is deleted [It]
[10]     /home/sebbonnet/gowork/src/github.com/sky-uk/core-operators/cassandra-operator/test/e2e/parallel/modification/modification_test.go:249
[10]
[10]     Timed out after 30.001s.
[10]     Expected event {Normal ClusterUpdate Custom config deleted for cluster test-cassandra-operator.mycluster-fdevi 2019-03-22 18:03:50.090478874 +0000 GMT m=+251.420548413}. Actual event: [{{ } {mycluster-fdevi.158e59c9b7eb31e1  test-cassandra-operator /api/v1/namespaces/test-
cassandra-operator/events/mycluster-fdevi.158e59c9b7eb31e1 69d314af-4ccc-11e9-88c4-0624fb7941ea 507389776 0 2019-03-22 18:00:44 +0000 GMT <nil> <nil> map[] map[] [] nil [] } 
[...]
[10]     /home/sebbonnet/gowork/src/github.com/sky-uk/core-operators/cassandra-operator/test/e2e/parallel/modification/modification_test.go:265
```
 ### Expected Behavior
The tests should consistently pass regardless of whether they are run in parallel or not.

 ### Steps to Reproduce
1. `cd $GOPATH/src/github.com/sky-uk/cassandra-operator/`
2. `make install-all`
3. `cd cassandra-operator`
4. `E2E_TEST=test/e2e/parallel/modification make e2e-test`

[...]

PR #130, opened 2019-06-18
Intermittent events loss

sebbonnet:
For #34


sebbonnet:
No need to keep this open now that investigating is complete

PR #137, opened 2019-06-20
Do not attempt updating cluster on snapshot change

### LINKED ISSUE #133: Move all defaulting to `v1alpha1helpers.SetDefaultsForCassandra`
In #126 we discovered that defaulting was happening too late. 
Cluster actions (e.g. `AddSnapshotCleanup`) were being calculated based on the un-defaulted Cassandra resource.
We've now put in place a new defaulting function `SetDefaultsForCassandra` which gets called immediately after the Cassandra object is received by our informer callbacks.
We now need to move *all* the defaulting from `cluster.New` to that function, including probe timeout defaults.

Later, as part of #71, we can also call that function from our webhook.

### LINKED ISSUE #134: Snapshot updates wrongly trigger a statefulset update
Adding, removing or updating Snapshot properties to a `cassandra` cluster definition triggers a cluster update operation even though no other cluster properties have changed. This results in patching the statefulset, which has no effect as far as the cluster is concerned as nothing is changed in the statefulet.

However this has 2 side-effects from the cassandra-operator point of view:
- it creates an unnecessary `WaitingForStatefulSetChange` event
- it waits `max(*readinessProbe.PeriodSeconds, 5)`, meaning the next cluster event is delayed by this much

We should change the behaviour so the above 2 operations do not take place.

sebbonnet:
Fix #133 and #134

* Apply defaults to both old and new cluster definition before any operations

When updating, adding or removing a Snapshot from a `cassandra` cluster definition a cluster update operation was triggered and resulted in patching the statefulsets even though pod specs should have not changed. On close inspection we've found this was due to the way we were applying defaults to `cassandra` definitions. In fact when both liveness and readiness probes were nil, defaults were being applied to the old `cassandra` definition but not the new one, which resulted in showing differences when there shouldn't have been any.

This PR fixes #134 by implementing #133 an so settings the defaults as early as possible before any cluster operations is carried out.
As part of this change optional `cassandra` fields are now directly accessed (without checking for nil pointer) as all optional fields would have been defaulted.

* Apply sidecar image changes

Updates the adjuster to ensure changes to the pod.SidecarImage property will trigger a cluster rolling update

* Handle error when patch cannot be created

Honours an error check when applying a patch for a statefulset


cassandra-operator/pkg/operator/operations/adjuster/adjuster.go
```diff
@@ -205,11 +220,7 @@ func (r *Adjuster) ensureChangeIsAllowed(oldCluster, newCluster *v1alpha1.Cassan
 }
 
 func (r *Adjuster) podSpecHasChanged(oldCluster, newCluster *v1alpha1.Cassandra) bool {
-	return !reflect.DeepEqual(oldCluster.Spec.Pod.CPU, newCluster.Spec.Pod.CPU) ||
-		!reflect.DeepEqual(oldCluster.Spec.Pod.Memory, newCluster.Spec.Pod.Memory) ||
-		!reflect.DeepEqual(oldCluster.Spec.Pod.LivenessProbe, newCluster.Spec.Pod.LivenessProbe) ||
-		!reflect.DeepEqual(oldCluster.Spec.Pod.ReadinessProbe, newCluster.Spec.Pod.ReadinessProbe) ||
-		!reflect.DeepEqual(oldCluster.Spec.Pod.BootstrapperImage, newCluster.Spec.Pod.BootstrapperImage)
+	return !reflect.DeepEqual(oldCluster.Spec.Pod, newCluster.Spec.Pod)
```
wallrj: We should probably consider using https://github.com/google/go-cmp here, as `resource.Quantity` implements `Equal` and I don't know whether you can simply compare to Quantity objects that are the same amount but with different SI units. See: https://github.com/sky-uk/cassandra-operator/issues/138

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -19,55 +37,6 @@ func NewControllerRef(c *v1alpha1.Cassandra) metav1.OwnerReference {
...
-	return v1alpha1.DefaultCassandraSnapshotImage
-}
-
-func GetDatacenter(c *v1alpha1.Cassandra) string {
-	if c.Spec.Datacenter == nil {
-		return v1alpha1.DefaultDCName
-	}
-	return *c.Spec.Datacenter
-}
-
```
wallrj: I'm glad we can get rid of these.

cassandra-operator/pkg/operator/operations/adjuster/adjuster.go
sebbonnet: hmm true. `DeepEqual` won't catch `resource.Quantity` equality. 
`resource.Quantity` doesn't implement `Equal`.
I think to make this work I'll have to implement a `Equal` at the `Pod` spec level

wallrj: According to the docs it does: https://godoc.org/k8s.io/apimachinery/pkg/api/resource#Quantity.Equal

sebbonnet: only made available 2months ago: https://github.com/kubernetes/apimachinery/blame/master/pkg/api/resource/quantity.go#L589

sebbonnet: but that's ok I can use `Cmp(y Quantity)` which is available in the version of apimachinery we use.

sebbonnet: I'll still need to implement an `Equal` though as our structs contains unexported fields that need to be explicitly allowed when doing the comparison - which is painful as we have alot of them

sebbonnet:
@wallrj I've added an `(p Pod) Equal(other Pod) bool` in the last commit so we can compare pod specs with different quantity amounts :)

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -11,6 +11,24 @@ import (
 	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/util/ptr"
 )
 
+var (
+	defaultLivenessProbe = v1alpha1.Probe{
```
alangibson01: I'm slightly troubled by the fact that these could end up being mutated if we accidentally end up using one of them without deep-copying. I know that might be unlikely, but given the world of pain you've gone through to get to this stage, any sliver of opportunity we leave open to that sort of issue is best to be closed as quickly as possible.

Because of that, could we have a couple of functions which return new default probe structs every time they're invoked? That way, we don't need to worry about always having to do `DeepCopy` before using them.

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -102,3 +87,56 @@ func setDefaultsForSnapshot(snapshot *v1alpha1.Snapshot) {
...
+		clusterDefinition.Spec.Pod.ReadinessProbe = defaultReadinessProbe.DeepCopy()
+	} else {
+		readinessProbe := clusterDefinition.Spec.Pod.ReadinessProbe
+		mergeProbeDefaults(readinessProbe, &defaultReadinessProbe)
+	}
+}
+
+func mergeProbeDefaults(configuredProbe *v1alpha1.Probe, defaultProbe *v1alpha1.Probe) {
+	if configuredProbe.TimeoutSeconds == nil {
+		configuredProbe.TimeoutSeconds = defaultProbe.TimeoutSeconds
```
alangibson01: This assignment (and the ones below) are copying pointers, which would potentially leave the way open for the value of the `defaultProbe` to be indirectly modified. (Imagine that later on there's some code that does `*configuredProbe.TimeoutSeconds = 99`.)

Now maybe this doesn't matter because `defaultProbe` is a throwaway object that won't be shared, but because that's just a convention and there's no compiler checking around it, there is potentially scope for something to go wrong. Do you think we need to worry about it?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -165,3 +166,16 @@ func (c *Cassandra) RackName(rack *Rack) string {
 func (c *Cassandra) CustomConfigMapName() string {
 	return fmt.Sprintf("%s-config", c.Name)
 }
+
+// Equal checks equality of two Pod. This is useful for testing with cmp.Equal
```
alangibson01:
```suggestion
// Equal checks equality of two Pods. This is useful for testing with cmp.Equal
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -165,3 +166,16 @@ func (c *Cassandra) RackName(rack *Rack) string {
...
+func (p Pod) Equal(other Pod) bool {
+	return &other != nil &&
```
alangibson01: Do we need this check? `other` is a struct, so `&other` will surely always return a non-`nil` value ...?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -165,3 +166,16 @@ func (c *Cassandra) RackName(rack *Rack) string {
...
+		reflect.DeepEqual(p.BootstrapperImage, other.BootstrapperImage) &&
+		reflect.DeepEqual(p.SidecarImage, other.SidecarImage) &&
+		reflect.DeepEqual(p.Image, other.Image) &&
+		reflect.DeepEqual(p.LivenessProbe, other.LivenessProbe) &&
+		reflect.DeepEqual(p.ReadinessProbe, other.ReadinessProbe) &&
+		(&p.StorageSize != nil && &other.StorageSize != nil && p.StorageSize.Cmp(other.StorageSize) == 0) &&
```
alangibson01: As above, `StorageSize` is defined on the struct as a value not a pointer, so a `nil` check should be redundant.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -165,3 +166,16 @@ func (c *Cassandra) RackName(rack *Rack) string {
...
+		(&p.Memory != nil && &other.Memory != nil && p.Memory.Cmp(other.Memory) == 0) &&
```
alangibson01: Same here.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -165,3 +166,16 @@ func (c *Cassandra) RackName(rack *Rack) string {
...
+		(&p.CPU != nil && &other.CPU != nil && p.CPU.Cmp(other.CPU) == 0)
```
alangibson01: And here.

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
sebbonnet: defaultProbe is meant to be a constant, so I agree we shouldn't pass it by value

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
sebbonnet: nope we don't
I'll also add a test for it

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
sebbonnet: you're right too.
It looks like my Java background is starting to show :)

sebbonnet:
@alangibson01 thanks for the review. It looks like my Java background is started to show :)

PR #132, opened 2019-06-18
Webhook based on controller-runtime

wallrj:
Required by #71 

Co-authored by: @kragniz

PR #139, opened 2019-06-20
WIP: Move validation to apis package and use apimachinery Validation helpers

wallrj:
Part of: #71

### DIFF
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
@@ -23,7 +23,10 @@ func ValidateCassandra(c *v1alpha1.Cassandra) field.ErrorList {
 }
 
 func validateCassandraSpec(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
-	return validateRacks(c, fldPath.Child("Racks"))
+	var allErrs field.ErrorList
+	allErrs = append(allErrs, validateRacks(c, fldPath.Child("Racks"))...)
+	allErrs = append(allErrs, validatePodResources(c, fldPath.Child("Pod"))...)
+	return allErrs
 }
 
 func validateRacks(clusterDefinition *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
@@ -84,6 +87,53 @@ func validateRacks(clusterDefinition *v1alpha1.Cassandra, fldPath *field.Path) f
 				),
 			)
 		}
+	}
+	return allErrs
+}
+
+func validatePodResources(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if c.Spec.Pod.Memory.IsZero() {
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				fldPath.Child("Memory"),
+				c.Spec.Pod.Memory.String(),
+				fmt.Sprintf(
+					"must be > 0 in Cassandra cluster %s.%s",
+					c.Namespace,
+					c.Name,
+				),
+			),
+		)
+	}
+	if !v1alpha1helpers.UseEmptyDir(c) && c.Spec.Pod.StorageSize.IsZero() {
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				fldPath.Child("StorageSize"),
+				c.Spec.Pod.StorageSize.String(),
+				fmt.Sprintf(
+					"must be > 0 when useEmptyDir is false for Cassandra cluster %s.%s",
+					c.Namespace,
+					c.Name,
+				),
+			),
+		)
+	}
+	if v1alpha1helpers.UseEmptyDir(c) && !c.Spec.Pod.StorageSize.IsZero() {
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				fldPath.Child("StorageSize"),
+				c.Spec.Pod.StorageSize.String(),
+				fmt.Sprintf(
+					"must be 0 when useEmptyDir is true for Cassandra cluster %s.%s",
+					c.Namespace,
+					c.Name,
+				),
+			),
+		)
 	}
 	return allErrs
 }
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
@@ -51,8 +51,7 @@ func validateRacks(clusterDefinition *v1alpha1.Cassandra, fldPath *field.Path) f
 				field.Invalid(
 					fldPath.Child("Replicas"),
 					rack.Replicas,
-					fmt.Sprintf("replicas must be a positive integer. Got value %d for Cassandra cluster definition: %s.%s",
-						rack.Replicas,
+					fmt.Sprintf("replicas must be a positive integer for Cassandra cluster: %s.%s",
 						clusterDefinition.Namespace,
 						clusterDefinition.Name,
 					),
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
@@ -1,0 +1,110 @@
+package validation
+
+import (
+	"fmt"
+	"reflect"
+
+	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1"
+	v1alpha1helpers "github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers"
+)
+
+func ValidateCassandra(c *v1alpha1.Cassandra) field.ErrorList {
+	allErrs := apimachineryvalidation.ValidateObjectMeta(
+		&c.ObjectMeta,
+		true,
+		apimachineryvalidation.NameIsDNSSubdomain,
+		field.NewPath("metadata"),
+	)
+	allErrs = append(allErrs, validateCassandraSpec(c, field.NewPath("spec"))...)
+	return allErrs
+}
+
+func validateCassandraSpec(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	return validateRacks(c, fldPath.Child("Racks"))
+}
+
+func validateRacks(clusterDefinition *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if len(clusterDefinition.Spec.Racks) == 0 {
+		allErrs = append(
+			allErrs,
+			field.Required(
+				fldPath,
+				fmt.Sprintf(
+					"no racks specified for cluster: %s.%s",
+					clusterDefinition.Namespace,
+					clusterDefinition.Name,
+				),
+			),
+		)
+		return allErrs
+	}
+
+	for i, rack := range clusterDefinition.Spec.Racks {
+		fldPath = fldPath.Child(fmt.Sprintf("%d:%s", i, rack.Name))
+		if rack.Replicas < 1 {
+			allErrs = append(
+				allErrs,
+				field.Invalid(
+					fldPath.Child("Replicas"),
+					rack.Replicas,
+					fmt.Sprintf("replicas must be a positive integer. Got value %d for Cassandra cluster definition: %s.%s",
+						rack.Replicas,
+						clusterDefinition.Namespace,
+						clusterDefinition.Name,
+					),
+				),
+			)
+		}
+		if rack.StorageClass == "" && !v1alpha1helpers.UseEmptyDir(clusterDefinition) {
+			allErrs = append(
+				allErrs,
+				field.Required(
+					fldPath.Child("StorageClass"),
+					fmt.Sprintf(
+						"either set useEmptyDir to true or specify storage class in Cassandra cluster %s.%s",
+						clusterDefinition.Namespace,
+						clusterDefinition.Name,
+					),
+				),
+			)
+		}
+		if rack.Zone == "" && !v1alpha1helpers.UseEmptyDir(clusterDefinition) {
+			allErrs = append(
+				allErrs,
+				field.Required(
+					fldPath.Child("StorageClass"),
+					fmt.Sprintf(
+						"either set useEmptyDir to true or specify zone in Cassandra cluster %s.%s",
+						clusterDefinition.Namespace,
+						clusterDefinition.Name,
+					),
+				),
+			)
+		}
+	}
+	return allErrs
+}
+
+func ValidateCassandraUpdate(old, new *v1alpha1.Cassandra) field.ErrorList {
+	allErrs := ValidateCassandra(new)
+
+	fldPath := field.NewPath("spec")
+	if !reflect.DeepEqual(new.Spec.Pod.Image, old.Spec.Pod.Image) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(
+				fldPath.Child("pod").Child("image"),
+				fmt.Sprintf(
+					"cannot change the image. "+
+						"old version: %s, new version: %s",
+					*old.Spec.Pod.Image, *new.Spec.Pod.Image,
+				),
+			),
+		)
+	}
+	return allErrs
+}
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kr/pretty"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1"
@@ -35,6 +36,7 @@ var _ = Describe("validation functions", func() {
 					Namespace: "ns1",
 				},
 				Spec: v1alpha1.CassandraSpec{
+					UseEmptyDir: ptr.Bool(false),
 					Racks: []v1alpha1.Rack{
 						{
 							Name:         "rack1",
@@ -42,6 +44,11 @@ var _ = Describe("validation functions", func() {
 							StorageClass: "fast",
 							Replicas:     1,
 						},
+					},
+					Pod: v1alpha1.Pod{
+						CPU:         resource.MustParse("0"),
+						Memory:      resource.MustParse("1Gi"),
+						StorageSize: resource.MustParse("100Gi"),
 					},
 				},
 			}
@@ -53,9 +60,7 @@ var _ = Describe("validation functions", func() {
 				err = validation.ValidateCassandra(cass).ToAggregate()
 				Expect(err).ToNot(HaveOccurred())
 			})
-			It("succeeds with a fully populated Cassandra object", func() {
-
-			})
+			It("succeeds with a fully populated Cassandra object", func() {})
 		})
 
 		Context("failure cases", func() {
@@ -92,6 +97,28 @@ var _ = Describe("validation functions", func() {
 						})
 						It("fails if Rack.Zone is empty", func() {
 							cass.Spec.Racks[0].Zone = ""
+						})
+					})
+				})
+
+				Context("Pod", func() {
+					It("fails if Memory is zero", func() {
+						cass.Spec.Pod.Memory.Set(0)
+					})
+					Context("UseEmptyDir=false", func() {
+						BeforeEach(func() {
+							cass.Spec.UseEmptyDir = ptr.Bool(false)
+						})
+						It("fails if StorageSize is zero", func() {
+							cass.Spec.Pod.StorageSize.Set(0)
+						})
+					})
+					Context("UseEmptyDir=true", func() {
+						BeforeEach(func() {
+							cass.Spec.UseEmptyDir = ptr.Bool(true)
+						})
+						It("fails if StorageSize is not zero", func() {
+							cass.Spec.Pod.StorageSize.Set(1)
 						})
 					})
 				})
--- a/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
+++ b/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
@@ -1,0 +1,118 @@
+package validation_test
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/kr/pretty"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1"
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation"
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/util/ptr"
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/test"
+)
+
+func TestCassandra(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecsWithDefaultAndCustomReporters(t, "Cluster Suite", test.CreateParallelReporters("cluster"))
+}
+
+func set(d interface{}, path string, value interface{}) {
+	segments := strings.Split(path, ".")
+	v := reflect.ValueOf(d)
+	for _, s := range segments {
+		for v.Kind() == reflect.Ptr {
+			v = v.Elem()
+		}
+		v = index(v, s)
+	}
+	v.Set(reflect.ValueOf(value))
+}
+
+func index(v reflect.Value, idx string) reflect.Value {
+	if i, err := strconv.Atoi(idx); err == nil {
+		return v.Index(i)
+	}
+	return v.FieldByName(idx)
+}
+
+type mutator func(*v1alpha1.Cassandra)
+
+func mutate(path string, value interface{}) mutator {
+	return func(c *v1alpha1.Cassandra) {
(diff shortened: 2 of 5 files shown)


cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -153,78 +145,6 @@ func (c *Cluster) Definition() *v1alpha1.Cassandra {
...
-		if retentionPeriodDays != nil && *retentionPeriodDays < 0 {
-			return fmt.Errorf("invalid snapshot retention policy retentionPeriodDays value %d, must be non-negative for Cassandra cluster definition: %s", *retentionPeriodDays, clusterDefinition.QualifiedName())
-		}
-
-		cleanupTimeoutSeconds := retentionPolicy.CleanupTimeoutSeconds
-		if cleanupTimeoutSeconds != nil && *cleanupTimeoutSeconds < 0 {
-			return fmt.Errorf("invalid snapshot retention policy cleanupTimeoutSeconds value %d, must be non-negative for Cassandra cluster definition: %s", *cleanupTimeoutSeconds, clusterDefinition.QualifiedName())
-		}
-
-		if retentionPolicy.CleanupSchedule != "" {
```
wallrj: I think this might have been a bug in the original. It allowed the `CleanupSchedule` to be empty.

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -980,9 +885,7 @@ var _ = Describe("creation of snapshot cleanup job", func() {
 	})
 
 	It("should not create a cleanup job if the retention policy is disabled in the cluster spec", func() {
-		clusterDef.Spec.Snapshot.RetentionPolicy = &v1alpha1.RetentionPolicy{
-			Enabled: ptr.Bool(false),
-		}
+		clusterDef.Spec.Snapshot.RetentionPolicy.Enabled = ptr.Bool(false)
```
wallrj: This test should have been failing, because it created a RetentionPolicy without a CleanupSchedule.

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -296,73 +267,7 @@ var _ = Describe("cluster construction", func() {
 			})
```
wallrj: I don't think this test does what it claims. There's nothing here that sets the operator --use-empty-dir flag, AFAICS.

wallrj: A good example of how code and tests might be simplified if we drop the UseEmptyDir support, in #74

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func validateRacks(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if len(c.Spec.Racks) == 0 {
+		allErrs = append(
+			allErrs,
+			field.Required(
+				fldPath,
+				"must not be empty",
```
wallrj: I decided to shorten the error messages; removed the qualified resource name from each message, because when these are displayed to the user by `kubectl` it will print out which cluster has failed the validation.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -0,0 +1,204 @@
...
+	v1alpha1helpers "github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers"
+)
+
+func ValidateCassandra(c *v1alpha1.Cassandra) field.ErrorList {
+	allErrs := apimachineryvalidation.ValidateObjectMeta(
+		&c.ObjectMeta,
+		true,
+		apimachineryvalidation.NameIsDNSSubdomain,
+		field.NewPath("metadata"),
+	)
```
wallrj: I'm not sure if it's necessary to validate the ObjectMeta.
As I understand it, the API server will reject a CRD that doesn't have the required Name and Namespace set.

wallrj: I put it here, because it's what we did in Navigator....but in that case we were using an aggregated API server so this check was probably necessary in that case. @munnerz is that right?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,261 @@
...
+			It("succeeds without a Snapshot.RetentionPolicy.CleanupTimeoutSeconds", func() {
+				cass.Spec.Snapshot.RetentionPolicy.CleanupTimeoutSeconds = nil
+			})
+		})
+
+		Context("failure cases", func() {
+			AfterEach(func() {
+				err = validation.ValidateCassandra(cass).ToAggregate()
+				fmt.Fprintf(GinkgoWriter, "INFO: Error message was: %s", err)
+				Expect(err).To(HaveOccurred())
```
wallrj: The original tests checked the error messages, but I found that a bit repetative, so I've instead opted to print out the error messages.
I probably should add a check which ensures  that the returned field.ErrorList contains an error for the expected field.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -145,6 +125,10 @@ func CopyInto(cluster *Cluster, clusterDefinition *v1alpha1.Cassandra) error {
 		snapshotImage := v1alpha1helpers.GetSnapshotImage(cluster.definition)
 		cluster.definition.Spec.Snapshot.Image = &snapshotImage
 	}
+
+	if err := validation.ValidateCassandra(cluster.definition).ToAggregate(); err != nil {
+		return err
+	}
```
wallrj: I considered moving this to the Informer callback functions, immediately after the defaulting, so that validation is performed before any actions are calculated.
I will do that once #137 is merged.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -153,108 +137,6 @@ func (c *Cluster) Definition() *v1alpha1.Cassandra {
...
-}
-
-func validateSnapshot(clusterDefinition *v1alpha1.Cassandra) error {
-	if clusterDefinition.Spec.Snapshot == nil {
-		return nil
-	}
-
-	if clusterDefinition.Spec.Snapshot.Schedule == "" {
-		return fmt.Errorf("no snapshot schedule property provided for Cassandra cluster definition: %s", clusterDefinition.QualifiedName())
-	}
```
wallrj: The new validation functions will return multiple errors rather than only the first error.

wallrj: And I found that `cron.Parse(` will return an error for empty schedule anyway, so I removed this extra check.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -153,108 +137,6 @@ func (c *Cluster) Definition() *v1alpha1.Cassandra {
...
-	return validateProbe("readiness", probe, clusterDefinition)
-}
-
-func validateProbe(name string, probe *v1alpha1.Probe, clusterDefinition *v1alpha1.Cassandra) error {
-	if probe.FailureThreshold == nil || *probe.FailureThreshold < 1 {
-		return fmt.Errorf("invalid failure threshold for %s probe, must be 1 or greater, got %d for Cassandra cluster definition: %s.%s", name, *probe.FailureThreshold, clusterDefinition.Namespace, clusterDefinition.Name)
-	}
-	if probe.InitialDelaySeconds == nil || *probe.InitialDelaySeconds < 1 {
-		return fmt.Errorf("invalid initial delay for %s probe, must be 1 or greater, got %d for Cassandra cluster definition: %s.%s", name, *probe.InitialDelaySeconds, clusterDefinition.Namespace, clusterDefinition.Name)
-	}
```
wallrj: Actually I think this can be zero. I changed it in the new implementation.
 * https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#probe-v1-core

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -50,31 +50,14 @@ var _ = Describe("cluster construction", func() {
 					RetentionPolicy: &v1alpha1.RetentionPolicy{
 						Enabled:             ptr.Bool(true),
 						RetentionPeriodDays: &retentionPeriod,
+						CleanupSchedule:     "0 9 * * *",
```
wallrj: The original validation allowed an empty CleanupSchedule, which I think is wrong.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
sebbonnet: do you have an example handy of what the kubectl output looks like?

wallrj: Here's how it looks when used in #132

```
kubectl -n default  apply -f examples/cassandra1.yaml 
Error from server ([spec.Racks.0:a.Replicas: Invalid value: 0: must be >= 1, spec.Pod.Memory: Invalid value: "0": must be > 0, spec.Pod.StorageSize: Invalid value: "0": must be > 0 when useEmptyDir is false]): error when creating "examples/cassandra1.yaml": admission webhook "vcass.core.sky.uk" denied the request: [spec.Racks.0:a.Replicas: Invalid value: 0: must be >= 1, spec.Pod.Memory: Invalid value: "0": must be > 0, spec.Pod.StorageSize: Invalid value: "0": must be > 0 when useEmptyDir is false]
```

And it turns out I'm wrong, it doesn't make it clear which cluster has the error. I'll see if I can improve that.

wallrj: And I'll see if I can remove the duplicate error messages.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -0,0 +1,195 @@
...
+			allErrs,
+			field.Invalid(
+				fldPath.Child("StorageSize"),
+				c.Spec.Pod.StorageSize.String(),
+				"must be 0 when useEmptyDir is true",
+			),
+		)
+	}
+	allErrs = append(allErrs, validateProbe(c, c.Spec.Pod.LivenessProbe, fldPath.Child("LivenessProbe"))...)
+	if *c.Spec.Pod.LivenessProbe.SuccessThreshold != 1 {
```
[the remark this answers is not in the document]
sebbonnet: Sorry, just realised that the old code used to have that logic too, but I can't find any explanatation for why that would be required.
I think we should remove it

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -0,0 +1,195 @@
...
+	allErrs = validateUnsignedInt(allErrs, c, fldPath.Child("TimeoutSeconds"), *probe.TimeoutSeconds, 1)
+	return allErrs
+}
+
+func validateSnapshot(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if c.Spec.Snapshot == nil {
+		return allErrs
+	}
+	if _, err := cron.Parse(c.Spec.Snapshot.Schedule); err != nil {
```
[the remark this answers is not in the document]
sebbonnet: we could also have a different message when Schedule is empty

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -0,0 +1,195 @@
...
+	allErrs = append(
+		allErrs,
+		validateSnapshotRetentionPolicy(c, fldPath.Child("RetentionPolicy"))...,
+	)
+	return allErrs
+}
+
+func validateSnapshotRetentionPolicy(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if c.Spec.Snapshot.RetentionPolicy == nil {
```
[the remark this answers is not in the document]
sebbonnet: I would do the nil check outside the function and only invoke it when required, like you've done for the other nillable fields -  e.g. `c.Spec.Snapshot.TimeoutSeconds `

cassandra-operator/pkg/cluster/cluster.go
sebbonnet: ok - #137 has now been merged - happy merging :)

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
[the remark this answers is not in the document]
wallrj: I thought it was weird too, but found that it's a requirement of the Kubernetes probe API:
>  successThreshold integer | Minimum  consecutive successes for the probe to be considered successful after  having failed. Defaults to 1. Must be 1 for liveness. Minimum value is  1.
-- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#probe-v1-core

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,261 @@
...
+			})
+			It("succeeds without a Snapshot.RetentionPolicy.RetentionPeriodDays", func() {
+				cass.Spec.Snapshot.RetentionPolicy.RetentionPeriodDays = nil
+			})
+			It("succeeds without a Snapshot.RetentionPolicy.CleanupTimeoutSeconds", func() {
+				cass.Spec.Snapshot.RetentionPolicy.CleanupTimeoutSeconds = nil
+			})
+		})
+
+		Context("failure cases", func() {
```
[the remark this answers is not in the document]
sebbonnet: Couldn't resist :) Have a look here: https://github.com/sky-uk/cassandra-operator/pull/140

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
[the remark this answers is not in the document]
wallrj: The empty case is already handled by the `cron.Parse` function:

```
validation functions ValidateCassandra failure cases Spec Snapshot RetentionPolicy 
  fails if CleanupSchedule is empty
  /home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go:250
INFO: Error message was: spec.Snapshot.RetentionPolicy.CleanupSchedule: Invalid value: "": is not a valid cron expression (Empty spec string)•
------------------------------
validation functions ValidateCassandra failure cases Spec Snapshot RetentionPolicy 
  fails if CleanupSchedule is not valid cron syntax
  /home/richard/go/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go:253
INFO: Error message was: spec.Snapshot.RetentionPolicy.CleanupSchedule: Invalid value: "x y z": is not a valid cron expression (Expected 5 to 6 fields, found 3: x y z)•

```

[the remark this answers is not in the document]
wallrj: So not much point in extracting to a separate function.

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -83,26 +82,6 @@ func NewWithoutValidation(clusterDefinition *v1alpha1.Cassandra) *Cluster {
...
-
-	if err := validateLivenessProbe(clusterDefinition.Spec.Pod.LivenessProbe, clusterDefinition); err != nil {
-		return err
-	}
-
-	if err := validateReadinessProbe(clusterDefinition.Spec.Pod.ReadinessProbe, clusterDefinition); err != nil {
-		return err
-	}
-
 	cluster.definition = clusterDefinition.DeepCopy()
```
wallrj: This method is now redundant, but I found various tests that rely on it. 
I'd like to defer cleaning this up for a followup branch.

cassandra-operator/pkg/operator/operator.go
```diff
@@ -174,78 +174,113 @@ func (o *Operator) safeGetConfigMap(obj interface{}) *v1.ConfigMap {
...
+			un.GetNamespace(),
+			un.GetName(),
+		)
 	}
 
 	return &c, nil
 }
 
-func logCassandraDecodeError(obj interface{}, err error) {
-	un := obj.(*unstructured.Unstructured)
```
wallrj: This was wrong. We'd already established that we didn't have an `unstructured.Unstructured` in `unstructuredToCassandra` who's error this was supposed to be logging.
I've tried to make the logging and error handling better in all the informer callbacks.

cassandra-operator/pkg/operator/operator.go
```diff
@@ -174,78 +174,113 @@ func (o *Operator) safeGetConfigMap(obj interface{}) *v1.ConfigMap {
...
+
 	o.eventDispatcher.Dispatch(&dispatcher.Event{Kind: operations.DeleteCluster, Key: clusterID, Data: clusterDefinition})
 }
 
+// clusterUpdated is called when there is an UPDATE operation on a Cassandra API object.
+// NB We only validate the *new* Cassandra object, not the old object.
+// This allows the operator to proceed if an invalid Cassandra API object is eventually corrected.
+// The operator will have ignored the invalid object when it was first added.
+// Webhook validation would ensure that the invalid Cassandra object is never accepted by the API server,
+// but we can't guarantee that a validating webhook has been deployed.
```
wallrj: This was highlighted by one of the modification e2e tests. 
See https://travis-ci.com/sky-uk/cassandra-operator/jobs/210753806#L6113

cassandra-operator/test/e2e/matchers.go
```diff
@@ -148,7 +148,7 @@ func (matcher *haveEvent) Match(actual interface{}) (success bool, err error) {
 }
 
 func (matcher *haveEvent) lastTimestampIsCloseTo(eventTimestamp v1.Time) bool {
-	return math.Abs(eventTimestamp.Sub(*matcher.expected.LastTimestampCloseTo).Seconds()) < 30
+	return math.Abs(eventTimestamp.Sub(*matcher.expected.LastTimestampCloseTo).Seconds()) < 45
```
wallrj: Fixes an intermittent E2E test timeout. See https://travis-ci.com/sky-uk/cassandra-operator/jobs/210725381#L7586

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
wallrj: I'm going to tackle this when we integrate the validation functions with the chosen webhook.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
munnerz: Correct. ObjectMeta doesn't need validation with CRD types as they already validate metadata 😄

wallrj: Thanks @munnerz. Now removed.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,374 @@
...
+			Entry(
+				"Rack.StorageClass must not be empty when UseEmptyDir=false",
+				"spec.Racks.rack1.StorageClass: Required value: because spec.useEmptyDir is true",
+				func(c *v1alpha1.Cassandra) {
+					c.Spec.UseEmptyDir = ptr.Bool(false)
+					c.Spec.Racks[0].StorageClass = ""
+				},
+			),
+			Entry(
+				"Rack.StorageClass must not be empty when UseEmptyDir=false",
```
sebbonnet: This is the same test case as immediately above, did you mean to have success case to check `StorageClass` is allowed to be empty when `UseEmptyDir=true`? Same goes for `Rack.Zone`

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,374 @@
...
+			Entry(
+				"Rack.Replicas must be >= 1",
+				"spec.Racks.rack1.Replicas: Invalid value: 0: must be >= 1",
+				func(c *v1alpha1.Cassandra) {
+					c.Spec.Racks[0].Replicas = 0
+				},
+			),
+			Entry(
+				"Rack.StorageClass must not be empty when UseEmptyDir=false",
+				"spec.Racks.rack1.StorageClass: Required value: because spec.useEmptyDir is true",
```
sebbonnet: error message should say `... spec.useEmptyDir is false`

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,374 @@
...
+				func(c *v1alpha1.Cassandra) {
+					c.Spec.UseEmptyDir = ptr.Bool(false)
+					c.Spec.Racks[0].StorageClass = ""
+				},
+			),
+			Entry(
+				"Rack.Zone must not be empty when UseEmptyDir=false",
+				"spec.Racks.rack1.Zone: Required value: because spec.useEmptyDir is true",
```
sebbonnet: error message should say `... spec.useEmptyDir is false`

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,374 @@
...
+				},
+			),
+			Entry(
+				"ReadinessProbe.SuccessThreshold must be >= 1",
+				"spec.Pod.ReadinessProbe.SuccessThreshold: Invalid value: 0: must be >= 1",
+				func(c *v1alpha1.Cassandra) {
+					c.Spec.Pod.ReadinessProbe.SuccessThreshold = ptr.Int32(0)
+				},
+			),
+			// Entry(
```
sebbonnet: commented out code can be removed

cassandra-operator/test/e2e/parallel/deletion/deletion_test.go
```diff
@@ -42,7 +43,7 @@ func createClustersInParallel(multipleNodeCluster, singleNodeCluster *TestCluste
 	AClusterWithName(multipleNodeCluster.Name).AndRacks(multipleNodeCluster.Racks).AndScheduledSnapshot(multipleNodeCluster.SnapshotConfig).IsDefined()
 	AClusterWithName(singleNodeCluster.Name).AndRacks(singleNodeCluster.Racks).AndScheduledSnapshot(singleNodeCluster.SnapshotConfig).IsDefined()
 
-	Eventually(PodReadyForCluster(Namespace, multipleNodeCluster.Name), 2*NodeStartDuration, CheckInterval).
+	Eventually(PodReadyForCluster(Namespace, multipleNodeCluster.Name), 3*NodeStartDuration, CheckInterval).
```
wallrj: This test fails intermittently, so I increased the timeout to match what we use in the creation tests.

sebbonnet: Changing the number of nodes to increase the timeout is confusing - I don't think that's a good idea.
If the the timeout is too short we should look at increasing it on a per node basis or add a buffer.

wallrj: :+1: Ah, I misunderstood the logic behind that.  So I need to increase  this in the Makefile.

sebbonnet: `NodeStartDuration` is defined in `test_environment`. 
It's set to 5m which seems pretty high already.
Perhaps there's something else going on here - if you add the failing build link I can take a look

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -0,0 +1,374 @@
...
+		)
+
+		DescribeTable(
+			"failure cases",
+			func(expectedMessage string, mutate func(*v1alpha1.Cassandra)) {
+				c := validCassandra()
+				mutate(c)
+				err := validation.ValidateCassandra(c).ToAggregate()
+				Expect(err).To(HaveOccurred())
+				fmt.Fprintf(GinkgoWriter, "INFO: Error message was: %q\n", err)
```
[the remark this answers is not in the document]
wallrj: It was useful for copying and pasting the expected error messages into the tests though.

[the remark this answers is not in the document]
wallrj: And it doesn't do a very good job if the strings are very different.
```
• Failure [0.000 seconds]
ValidateCassandra failure cases [It] Rack.StorageClass must be empty when UseEmptyDir=true 
/home/richard/go/pkg/mod/github.com/onsi/ginkgo@v1.8.0/extensions/table/table_entry.go:46

  Expected
      <string>: "...spec.Pod.St..."
  to equal               |
      <string>: "...spec.Racks...."

```

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
wallrj: Added those success cases.

cassandra-operator/test/e2e/parallel/deletion/deletion_test.go
wallrj: Here's the failing test: https://travis-ci.com/sky-uk/cassandra-operator/jobs/211057857#L5633

I've reverted this change.

wallrj:
I pushed a few more commits which I think should fix the E2E test failure.
Also added Job diagnostic output and increased the default POD timeout by 30s as discussed.

wallrj:
I'd like to improve this https://github.com/sky-uk/cassandra-operator/blob/master/cassandra-operator/test/e2e/parallel/snapshot/snapshot_test.go#L93

It seems to me that `HaveJobSpec` ought to be given a job name, and only match the spec against the job with that name, else fail. What do you think @sebbonnet

sebbonnet:
I agree, the matchers assertions makes it difficult to understand failed cases.

sebbonnet:
I mean I would keep the existing properties in `JobExpectation`, but add a JobName which should be the first thing checked in `HaveJobSpec`. We need to keep the other checks as the operator only cares about scheduling a job with the right properties, not that the job was triggered successfully - this is the job of cassandra-snapshot.

wallrj:
I'm pretty sick of this branch now, so I suggest making the Matcher changes in a later PR. I created https://github.com/sky-uk/cassandra-operator/issues/146 to remind us.

PR #140, opened 2019-06-21
Use describe table to shorten tests

sebbonnet:


cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
```diff
@@ -17,196 +18,91 @@ func TestTypes(t *testing.T) {
...
+			PeriodSeconds:       ptr.Int32(2),
+			InitialDelaySeconds: ptr.Int32(3),
+			FailureThreshold:    ptr.Int32(4),
+			TimeoutSeconds:      ptr.Int32(5),
+		},
+	}
+	pod2 := pod1.DeepCopy()
+
+	podToChange := pod1.DeepCopy()
+	applyChange(podToChange)
```
sebbonnet: As `BeforeEach` cannot be used between `Entry` in `DescribeTable`, we reinitialise the pods for every `Entry` inside the body function. See https://github.com/onsi/ginkgo/issues/378

PR #144, opened 2019-06-27
A comparison of webhook libraries

wallrj:
Required by #71 

See:
 * controller-runtime webhook spike: #132 
 * openshift webhook spike: #87

### DIFF
--- a/docs/enhancements/71-webhook.md
+++ b/docs/enhancements/71-webhook.md
@@ -1,0 +1,62 @@
+# Comparison of ﻿Webhook Libraries
+
+We have evaluated two webhook libraries:
+
+1. Generic Admission Server: https://github.com/openshift/generic-admission-server
+2. Controller Runtime: https://github.com/kubernetes-sigs/controller-runtime
+
+Here is our evaluation / comparison:
+
+## Popularity
+
+Openshift
+* Used by 10-20 projects
+* https://github.com/search?l=Go&q=%22generic-admission-server%22&type=Code
+Controller tools
+* Used by 5-10  projects
+* https://github.com/search?q=%22%2Bkubebuilder%3Awebhook%22&type=Code
+
+
+## Stability
+
+Controller-tools:
+* API still in flux:
+   * for example: https://github.com/kubernetes-sigs/controller-runtime/pull/497
+
+Openshift:
+* Established October 2017.
+* Developed by Redhat and used throughout Openshift.
+* Stable.
+
+
+## Maintainability
+
+Controller tools:
+1. Currently unstable.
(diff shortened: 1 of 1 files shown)


sebbonnet:
Required by #71

docs/enhancements/71-webhook.md
```diff
@@ -0,0 +1,126 @@
...
+
+Openshift:
+
+The Openshift library makes it simple to also have the webhook server authenticate the requests it receives from the Kubernetes API servers.
+It takes advantage of the well established mutual authentication mechanism for aggregate API servers. See https://github.com/openshift/generic-admission-server#architecture
+
+Additionally the Openshift library performs authorization of incoming requests.
+It POSTs SubjectAccessReview API resources to the Kubernetes API server,
+asking whether the user who made the original API request is permitted to interact with the webhook server endpoint.
+For this reason, the webhook service account needs a role binding to the `system:auth-delegator` cluster role.
```
sebbonnet: According to https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/, an additional `extension-apiserver-authentication-reader` role is also required for accessing the `extension-apiserver-authentication` configmap.

sebbonnet: According to https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#kubernetes-apiserver-client-authentication, there are additional flags required to allow the apiserver to communicate with the extended api-server. Namely `proxy-client-*-file` and `--requestheader-*`.

wallrj: I think those flags are configure by default by most depllyment tools: https://kubernetes.io/docs/setup/best-practices/certificates/#all-certificates

wallrj:
I've added a section on Testability and a new item on the API server flags required by the Openshift aggregate API server approach.

sebbonnet:
I've had some more feedback from @balooo who thought that Openshift was depicted in this doc as the better solution. This made me think that we're probably not explaining clearly enough that the 2 libraries implements different types of webhooks and that this is really what the comparison should be about.  

In my opinion the choice of library is less important than the webhook model we want to follow.
To me the choice is between a simple http server webhook and an extended API server.
I feel that the former is, albeit less powerful, a well known concept and would be easier for users to reason about / setup and so to adapt. Besides it is already used in the open SDK and is likely to gain more momentum. The unstability of the library is not a huge concern as there isn't a great deal of features we use and the library is fundamentally a thin wrapper around an https server.
For these reasons my choice would go for the simple http server and so the controller-runtime library.

Issue #151, opened 2019-07-03
Intermittent licence-compliance-checker failures


wallrj:
```
cassandra-operator/cassandra-sidecar $ make build
...
ERRO[0005] Some licences are not compliant and/or cannot be identified: restricted: [{/home/richard/go/pkg/mod/github.com/hashicorp/hcl@v1.0.0 [{MPL-2.0-no-copyleft-exception 0.9874513} {MPL-2.0 0.9874513}] }], unidentifiable: [] 
make: *** [Makefile:50: validate-licences] Error 1
```

I don't understand why this isn't occurring on travis.
And it occurs intermittently on my laptop.

The licence checker prints two possible licenses, with equal probabilities:
`[{MPL-2.0-no-copyleft-exception 0.9874513} {MPL-2.0 0.9874513}] `

Perhaps the license matching code is non-deterministic.

sebbonnet:
> Perhaps the license matching code is non-deterministic.

True. I think that might actually be a bug in SPDX database as both MPL 2.0 and MPL 2.0 no copy licences text are the same. I raised this for it, as unsure if this is intentional behaviour: https://github.com/spdx/LicenseListPublisher/issues/55

I also raised https://github.com/sky-uk/licence-compliance-checker/issues/20 to make our licence check more deterministic - fixed is in https://github.com/sky-uk/licence-compliance-checker/pull/21

PR #152, opened 2019-07-03
Downgrade to cobra 0.0.3 to get rid of our dependency on hashicorp/hcl

wallrj:
Work around for: #151


sebbonnet:
This won't be needed after https://github.com/sky-uk/licence-compliance-checker/pull/21 is merged.
You can also override the licence for the hcl project using this flag in the meantime: `-m github.com/hashicorp/hcl=MPL-2.0`

sebbonnet:
superceded by #154

Issue #148, opened 2019-07-03
Move cluster update validation to the validation package


wallrj:
In #139 we added a validation package for validating new Cassandra clusters.
Next we need to add functions to validate cluster updates.
Dissallow all updates by default and allow only the modifications that we support in the operator.

sebbonnet:
Invalid cluster updates are rejected in https://github.com/sky-uk/cassandra-operator/blob/v0.70.1-alpha.3/cassandra-operator/pkg/operator/operations/adjuster/adjuster.go#L199, so this work would look into exposing this validation so that it can also be invoked in the webhook on updates

PR #156, opened 2019-07-10
Move update validation to the validation package

wallrj:
Fixes: #148 

I tried to implement this as a function that knows a list of "allowed" changes and rejects everything else, but got bogged down with the testing.
So here's a very basic implementation instead....for now.

I notice that there are certain changes which are allowed by the current implementation which perhaps should be forbidden. Such as:
 * Change Pod Bootstrapper image
 * Change sidecar image
 * Change StorageSize
 * Change Snapshot image.


wallrj:
Ah, I see now that there's another function that catches forbidden changes. `Adjuster.ensureChangeIsAllowed`  allowed rack deletion, but that is then forbidden later on, in `UpdateClusterOperation.Execute`...and it also logs an Event.

sebbonnet:
Scaling down operations have not yet been implemented, so we need to reject them.
I also think we should record an event in case of a validation failure be it for adding or updating a cluster. Operator logs are useful, but users ought to be able to rely on kubernetes events to diagnose cluster changes. 
We had this note in #40 which is a long term goal, but it's good pratice to add missings events as we go along:
> We should make sure that each operation supported by the operator is associated with an event, so cluster admins can follow the entire cluster lifecycle using kubernetes events.
The goal should be that cluster admins don't ever need to look in the operator logs to see any information about their cluster lifecycle. The logs should be for our use, to identify and resolve issues.

wallrj:
Ok. I've now added validation to forbid rack scale-in and rack deletion.
Also removed all the scale-in and rack deletion code  and tests from the adjuster package since it won't be e2e tested any more.
We can add it back when we implement those operations properly.

wallrj:
Looks like build failure was caused by a bug in controller-tools https://github.com/sky-uk/cassandra-operator/issues/158

PR #157, opened 2019-07-10
Clarify webhook concept implemented and add conclusion

sebbonnet:
Explain that the 2 libraries implement different types of webhooks and conclude that the choice of webhook model is the determining factor.

Also moves the design documents under `docs`.

Required by #71


sebbonnet:
I'll pick this up again after the reconciliation work :)

sebbonnet:
Closing as this doc is stale and not worth refreshing for now.
The doc is removed algether in https://github.com/sky-uk/cassandra-operator/pull/246

Issue #162, opened 2019-07-31
Reconciler prevents update if operator is redeployed


sebbonnet:
When a cluster is created and the operator subsequently redeployed, updates to the Cassandra definition are never invoked given that the reconciliation logic exits early with this error:
```
{"level":"error","ts":1564593098.1848216,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"cassandra","request":"test-cassandra-operator/small-cluster","error":"couldn't find a previousCassandra","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/sebbonnet/gowork/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/sebbonnet/gowork/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0-beta.3/pkg/internal/controller/controller.go:218\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/sebbonnet/gowork/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0-beta.3/pkg/internal/controller/controller.go:192\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/sebbonnet/gowork/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0-beta.3/pkg/internal/controller/controller.go:171\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/sebbonnet/gowork/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/sebbonnet/gowork/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/sebbonnet/gowork/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:88"}
```

This issue was introduced in #150 which is not yet released, and so only affects master.

sebbonnet:
Fix in #170

Issue #160, opened 2019-07-30
Duplicate configmap volume created during scale up


sebbonnet:
Investigate build failure on merge to master of #150 
This could be a bug introduced by the new reconciling mechanism?

`Allowable cluster modifications > should create a new stateful set when a new rack is added to the cluster definition` failed to update the cluster

**cassandra-operator logs**
(1563449597 is 11:33:17)
```
test-cassandra-operator/mycluster-dy93o
{"level":"info","ts":1563449607.335081,"logger":"cassandra-operator.reconciler","msg":"got a new configmap","request":"test-cassandra-operator/mycluster-dy93o","configMap":{"apiVersion":"v1","kind":"ConfigMap","namespace":"test-cassandra-operator","name":"mycluster-dy93o-config"}}
time="2019-07-18T11:33:28Z" level=info msg="stateful set test-cassandra-operator.mycluster-dy93o-a is ready" source="accessor.go:126"
time="2019-07-18T11:33:28Z" level=info msg="Event will trigger 1 operations" key=test-cassandra-operator.mycluster-dy93o type=ADD_CUSTOM_CONFIG
time="2019-07-18T11:33:28Z" level=info msg="Applying update for rack a in cluster test-cassandra-operator.mycluster-dy93o"
test-cassandra-operator/mycluster-dy93o
```

**Kubernetes Events**
Events suggests custom config volume is duplicated
```
 Warning  FailedCreate      6s (x17 over 1m)  statefulset-controller  create Pod mycluster-dy93o-a-0 in StatefulSet mycluster-dy93o-a failed error: Pod "mycluster-dy93o-a-0" is invalid: [spec.volumes[4].name: Duplicate value: "cassandra-custom-config-mycluster-dy93o", spec.initContainers[1].volumeMounts[3].mountPath: Invalid value: "/custom-config": must be unique]
```

**Cluster definition**
There is nothing out of the ordinary in the cassandra definition. 
```
==== Cluster mycluster-dy93o =====
No resources found.

==== Cassandra definition =====
apiVersion: core.sky.uk/v1alpha1
kind: Cassandra
metadata:
  annotations:
    reconciled.cassandra.core.sky.uk: "true"
  clusterName: ""
  creationTimestamp: 2019-07-18T11:33:17Z
  generation: 1
  name: mycluster-dy93o
  namespace: test-cassandra-operator
  resourceVersion: "3756"
[...]

PR #150, opened 2019-07-03
controller-runtime based controller

kragniz:
Required by #53

Issue #138, opened 2019-06-20
Consider using https://github.com/google/go-cmp instead of reflect.DeepEqual


wallrj:
DeepEqual isn't always able to compare complex data types.
`resource.Quantity` and `time.Time` for example.

See:
 * https://github.com/google/go-cmp

sebbonnet:
Since I've used `go-cmp` as part of #137 when comparing the pod.Spec, I think the only other occurrence where this might be useful is here: https://github.com/sky-uk/cassandra-operator/blob/v0.70.1-alpha.3/cassandra-operator/pkg/operator/operator.go#L243

I think this is a nice to have though, as we have no `time.Time` or `resource.Quantity` that live outside the `Pod` spec which we already compare correctly. Having the top level struct compared with `go-cmp` would allow to close the gap and ensure any future properties added outside of `Pod` spec will be compared correctly.

PR #164, opened 2019-08-14
WIP - Reconcile by reconstructing the current state 

sebbonnet:
This is a follow up to #150 
This changes the reconciliation logic to reconstruct the Cassandra definition from the current state instead of tracking the previous resources state in memory which doesn't survive operator redeployments.  
This approach allows for resources to be reconciled when the operator is restarted and eventually when the resources are modified outside of the control of the operator (e.g. editing a statefulset, unintentionally deleted a job).

At a high level, the sequence of events is as follows:
1. the manager watches resources
2. on change a reconcile request is sent to the reconciler that process them 1 at a time 
3. the reconciler reconstructs the current state by interrogating the api server (most of the work in this PR)
4. current and desired state are compared to determine the type of changes to apply
5. the corresponding event is dispatched asynchronously and the reconciliation request completes
6. the event is processed by the receiver which calculates the delta and apply the changes

There are a few problems with this approach:
A. rebuilding the current state is hard and may complicate things when upgrading the operator
B. processing the events asynchronously doesn't give the chance for the reconciler to retry requests when the delta changes cannot be applied for whatever reason (e.g. api server down or resource creation / update failing)

Regarding A, perhaps a nicer approach would be to advance the current state to the desired state, without trying to work out the delta - e.g. look up existing resource an add all the parts from the desired state (regardless of whether the state is already in sync)
Regarding B, removing the dispatcher sounds reasonable at first, but this would mean reconcile requests could take a long time to complete and significantly delay updates for other clusters. Increasing the number of work queue items (MaxConcurrentReconciles) would allow concurrent changes, but won't stop parallel updates for the same cluster which could cause disruption (e.g. multiple node down when doing rolling updates) or leave it in a unknown state. Implementing a custom reconciler with a work queue per cluster would solve that problem. 

This PR is raised for visibility and fixes the scenario described in #162.
It also addresses #160 and #138 which I plan on splitting into their own PRs.
There are also a few tests marked pending that show missing behaviours - e.g. ensuring changes made outside of the controller are reconciled, ensure existing clusters are known to the operator on restart (even when no change was detected)

Required by #53


sebbonnet:
Closed in favour of #170.
Noteworthy, the test proving statefulset, jobs (and eventually headless service) are resurrrected has not been included in the new PR: https://github.com/sky-uk/cassandra-operator/pull/164/commits/955dbc269241aedc404ab1dc697f5c5c54297696#diff-9458ab2b0334d93579895377c6d0b63e

PR #166, opened 2019-08-15
Use cmd.Equal to compare CassandraSpec

### LINKED ISSUE #138: Consider using https://github.com/google/go-cmp instead of reflect.DeepEqual

sebbonnet:
As reflect.DeepEqual does not fit for the following reasons:
- order of keyspaces and racks are not important
- resourceQuantity comparison cannot rely on the raw value
- comparison is different depending on whether UseEmptyDir is set

Closes #138


sebbonnet:
Cherry picked from #164 (59b294abaf5107153fbd45ddaf04b816de6ef614 & beed790d54970aae07b2881a1f1b2b9fe3c978fd)

sebbonnet:
Also cherry picked cfc13a0 as the custom equality check is no longer required once #168 is merged

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
```diff
@@ -63,18 +63,184 @@ var _ = Describe("Cassandra Types", func() {
...
+		Snapshot:    snapshotSpec(),
+	}
+	cassandra2 := cassandra1.DeepCopy()
+
+	applyChange(cassandra1)
+
+	Expect(expectCheck(cassandra1, cassandra2)).To(BeTrue())
+}
+
+func cassandraUsingEmptyDirComparisonCheck(applyChange func(cassandra *CassandraSpec), expectCheck func(cassandra, otherCassandraSpec *CassandraSpec) bool) {
```
alangibson01: Unused function - are there some missing tests that this was put in place for?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
```diff
@@ -63,18 +63,184 @@ var _ = Describe("Cassandra Types", func() {
...
+	}
+	cassandra1.UseEmptyDir = ptr.Bool(true)
+	cassandra2 := cassandra1.DeepCopy()
+
+	applyChange(cassandra1)
+
+	Expect(expectCheck(cassandra1, cassandra2)).To(BeTrue())
+}
+
+func snapshotEqualityCheck(applyChange func(snapshot *Snapshot)) {
```
alangibson01: A minor naming quibble - I think I'd prefer something like `mutation` or `mutator`, rather than `applyChange`. What do you think?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
```diff
@@ -63,18 +63,184 @@ var _ = Describe("Cassandra Types", func() {
...
+		)
+		DescribeTable("inequality",
+			cassandraInequalityCheck,
+			Entry("when one cassandra has a different datacenter", func(cass *CassandraSpec) { cass.Datacenter = ptr.String("otherDc") }),
+			Entry("when one cassandra has empty dir enabled", func(cass *CassandraSpec) { cass.UseEmptyDir = ptr.Bool(true) }),
+			Entry("when one cassandra has no racks", func(cass *CassandraSpec) { cass.Racks = nil }),
+			Entry("when one cassandra has an empty racks list", func(cass *CassandraSpec) { cass.Racks = []Rack{} }),
+			Entry("when one cassandra has a different number of racks", func(cass *CassandraSpec) { cass.Racks = []Rack{*rackSpec("a")} }),
+			Entry("when one cassandra has a no snapshot", func(cass *CassandraSpec) { cass.Snapshot = nil }),
+			Entry("when one cassandra has a different snapshot", func(cass *CassandraSpec) { cass.Snapshot.Schedule = "1 2 3 4" }),
```
alangibson01:
```suggestion
			Entry("when one cassandra has a different schedule", func(cass *CassandraSpec) { cass.Snapshot.Schedule = "1 2 3 4" }),
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
```diff
@@ -63,18 +63,184 @@ var _ = Describe("Cassandra Types", func() {
...
+		DescribeTable("equality",
+			cassandraEqualityCheck,
+			Entry("if all fields are equal", func(cass *CassandraSpec) {}),
+			Entry("when the racks are in a different order", func(cass *CassandraSpec) { cass.Racks = []Rack{*rackSpec("b"), *rackSpec("a")} }),
+		)
+		DescribeTable("inequality",
+			cassandraInequalityCheck,
+			Entry("when one cassandra has a different datacenter", func(cass *CassandraSpec) { cass.Datacenter = ptr.String("otherDc") }),
+			Entry("when one cassandra has empty dir enabled", func(cass *CassandraSpec) { cass.UseEmptyDir = ptr.Bool(true) }),
+			Entry("when one cassandra has no racks", func(cass *CassandraSpec) { cass.Racks = nil }),
```
alangibson01: Is there any meaningful distinction between this test and the following one?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
sebbonnet: Yes, there was and you can see the removal here: https://github.com/sky-uk/cassandra-operator/pull/164/commits/cfc13a036b01d8c912c5148b49d84cb7c4dec80e
This leftover is due to a bad merge, I have too many branches!! :laughing: 

I used to have logic in place that would compare the racks differently depending on whether useEmptyDir was set and ignore StorageClass and Zone when it was since the properties are not used. 
I then figured that it's better to actively rejects these properties upfront in #168 and so decided to remove that piece of logic.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types_test.go
sebbonnet: no, not really as DeepEquals already deals with `nil` value by initialising the `nil` interface - see https://github.com/golang/go/blob/release-branch.go1.12/src/reflect/value.go#L2254
I thought it is best to have an explicit check to make sure our implementation will catch these scenarios.
I've done the same checks for nil and empty keyspaces too

PR #168, opened 2019-08-15
Fix affinity rules for emptydir

### LINKED ISSUE #167: Unable to create a cluster using empty without specifying storage
Storage properties are not required when using emptyDir volumes and so it should be possible to create the following cluster definition:
```
kind: Cassandra
metadata:
  name: mycluster
spec:
  useEmptyDir: true
  pod:
    image: "cassandra:3.11"
    cpu: "0"
    memory: 1Gi
    storageSize: "0"
  racks:
  - name: a
    replicas: 1
    storageClass: ""
    zone: ""
```

Yet when attempting to create the above cluster, the pods cannot be scheduled as the nodeAffinity rule is broken, with no value for the zone:
```
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: failure-domain.beta.kubernetes.io/zone
            operator: In
            values:
            - ""
``` 

Ideally we would want to remove the `useEmptyDir` properties and replace it by defining an emptyDir type of storage rather than a PV type. This idea was somewhat suggested here already: see https://github.com/sky-uk/cassandra-operator/issues/74

sebbonnet:
Fixes #167 

Also actively rejects invalid storage attributes in a cluster definition when using emptyDir such as Zone and StorageClass


sebbonnet:
Cherry picked 7002e7a, 0ef9ea1 and ba5f6d4 from #164

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -56,6 +56,16 @@ func validateRacks(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
 				),
 			)
 		}
+		if rack.StorageClass != "" && useEmptyDir {
```
alangibson01: Is there a case for making `StorageClass` an optional property, then?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -65,6 +75,16 @@ func validateRacks(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
 				),
 			)
 		}
+		if rack.Zone != "" && useEmptyDir {
```
alangibson01: As above. Also, do we need to make sure that if one of `StorageClass` or `Zone` is set that the other one is as well? Or is that already covered by other validation?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -56,6 +56,16 @@ func validateRacks(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
 				),
```
alangibson01: IntelliJ claims that the `aValidCassandraUsingEmptyDir` and `makeCompatibleWithUsingEmptyDir` functions is this file are unused.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
sebbonnet: Possibly, but I think we need to take a step back and review the way we expose storage. We already have stories to review the storage #78 and empty dir in #74

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
sebbonnet: Short answer is we don't need to.
The logic is derived based on whether useEmptyDir is set or not. We then scan all the racks and make sure they comply with the rules attached to useEmptyDir.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
sebbonnet: both functions are used in the test file `cassandra_test.go`, there isn't any refs in `cassandra.go`

alangibson01:
Happy for you to merge this now, but just wondering if `aValidCassandraUsingEmptyDir` and `makeCompatibleWithUsingEmptyDir` should be in the file where they're used.

sebbonnet:
@alangibson01 
> Happy for you to merge this now, but just wondering if aValidCassandraUsingEmptyDir and makeCompatibleWithUsingEmptyDir should be in the file where they're used.

There are in `cassandra_test.go` and used there. Am I missing something?

Issue #53, opened 2019-04-03
Reconcile cluster state based on its definition at regular intervals


sebbonnet:
### Is your feature request related to a problem? Please describe
There is no reconciliation between the cluster definition and the actual state of the cluster.
This is problematic because:
 - (at best) it could lead to some confusion
 - it could create issues where the next update to the cluster definition cannot modify the cluster as its state is incompatible or will only partially modify the cluster leaving it in an inconsistent state (e.g. configmap updates will not be propagated to manually added rack).
 - Kubernetes resources that are mistakenly deleted would need to be manually recreated, which could be both tedious and prone to error
  - it deviates from the Kubernetes model, where resource definitions are the golden source 
  - updates that fail for whatever reason (e.g. kubernetes apiserver unavailable) are not reapplied, potentially leaving the cluster in an inconsistent state

### Describe the solution you'd like
Move to a model where each Kubernetes resource created by the operator is reconciled with the cluster definition intended state. This would: 
- ensure the cluster definition remains the source of truth and so remains true to the kubernetes philosophy
 - ensure cluster updates that fail for whatever reasons (e.g. kubernetes apiserver unavailable) will eventually be applied, therefore avoiding the need to write complicated retry logic

### Tasks
- [x] Introduce reconciliation hook via controller-runtime based controller - #150
  - leftovers of #150: refactor stuff out of cmd package, remove duplicate status.go and operator.go, fix logs timestamp, confirm RBAC changes to allow watch for secrets and pods, configmaps watch already allowed and to allow update on cassandra - #175
- [x] Add reconciliation logic for changes while operator is down - #171
- [x] Add reconciliation logic for changes that were partially applied - #171 
- [x] Add reconciliation logic for manually changes to resources managed by the operator - #171 
- [x] Add reconciliation logic for headless service changes - #171 
- [x] Expose reconciliation syncPeriod as cmd line arg - #172
- ~Make it possible to pause the reconciliation~ - descope, as it's not clear what scenario would benefit from this

sebbonnet:
We want the reconciliation to recover from cases where changes were not applied (e.g. apiserver down) or while the operator was down.

**Use Cases**
While the operator down:
- cluster is added / deleted
- replica is added
- rack is added
- pod properties (e.g. mem, cpu, probes) are modified
- snapshot is added / removed / modified
- snapshot cleanup is added / removed / modified
- configmap is added / removed / modified

We also need to cover cases where the changes were partially applied, for instance:
[...]

PR #170, opened 2019-08-16
Reconcile by determining current state

sebbonnet:
This is a follow up to #150 and supercedes #164
This changes the reconciliation logic to reconstruct the Cassandra definition from the current state instead of tracking the previous resources state in memory which doesn't survive operator redeployments.

This approach allows for resources to be reconciled when the operator is restarted and _eventually_ when the resources are modified outside of the control of the operator (e.g. editing a statefulset, unintentionally deleted a job). I say _eventually_ because at the moment each change is applied asynchronously via our `dispatcher`. As a result, changes that fail the first time may not be applied before the next reconciliation loop (either at next change or at worst at the next `manager.Options.SyncPeriod` which defaults to 10hrs). Follow-up work will be done to address that, perhaps by swapping event dispatching and reconciliation, reducing the `SyncPeriod`, improving the changes application or a combination of that. 

As part of this change, the list of clusters is now managed by the `reconciler`
The list of clusters is mainly used to allow retrieval of metrics from
clusters managed by the operator.
Previously new clusters where only added on creation and on updates
which did not work for scenarios where the operator was restarted
but nothing needed to be reconciled.
This changes the logic to register new cluster definition at each
reconciliation, so the list is correct even when nothing is changed.

Required by #53


cassandra-operator/pkg/cluster/state.go
```diff
@@ -0,0 +1,410 @@
+package cluster
```
alangibson01: I'm finding it hard to follow the content of what's in this file. Could we split it up into a couple of files, each with a couple of structs or interfaces in it? Some extra comments on what each of the types is supposed to do would be helpful as well.

cassandra-operator/test/e2e/pod_events.go
```diff
@@ -57,24 +57,39 @@ func watchEvents(watcher watch.Interface, events eventLog) {
...
-		lastCreationTimeForFirstPod, err := e.lastCreationTimeForPod(firstPod)
-		if err != nil {
-			return false, err
-		}
-
-		deletionTimeForSecondPod, err := e.deletionTimeForPod(secondPod)
-		if err != nil {
-			return false, err
-		}
+func (e *PodEventLog) PodsRecreatedOneAfterTheOther(pod, otherPod string) (bool, error) {
```
alangibson01: I think the naming of the arguments needs to be clearer. Which of the pods is the one we expect to be recreated first?

cassandra-operator/test/e2e/pod_events.go
```diff
@@ -57,24 +57,39 @@ func watchEvents(watcher watch.Interface, events eventLog) {
...
+	}
+	deletionTimeForOtherPod, err := e.deletionTimeForPod(otherPod)
+	if err != nil {
+		return false, err
+	}
 
-		if deletionTimeForSecondPod.Before(lastCreationTimeForFirstPod) {
-			return false, fmt.Errorf("second pod was deleted before first pod was recreated. Second pod deleted at: %v, first pod last created at: %v", deletionTimeForSecondPod, lastCreationTimeForFirstPod)
-		}
+	if deletionTimeForPod.Before(deletionTimeForOtherPod) && !lastCreationTimeForPod.Before(deletionTimeForOtherPod) {
```
alangibson01: I don't understand the logic of this check and the following one. Can we talk through it?

cassandra-operator/cmd/operator/controller.go
```diff
@@ -17,118 +19,212 @@ import (
...
-		r.eventDispatcher.Dispatch(&dispatcher.Event{Kind: operations.DeleteCluster, Key: clusterID, Data: deletedCass})
-		delete(r.previousCassandras, request.NamespacedName.String())
+		logger.Debug("Cassandra definition not found. Going to delete associated resources")
+		cassandraToDelete := &v1alpha1.Cassandra{ObjectMeta: metav1.ObjectMeta{Name: request.Name, Namespace: request.Namespace}}
+		r.eventDispatcher.Dispatch(&dispatcher.Event{Kind: operations.DeleteCluster, Key: ctx.eventKey, Data: cassandraToDelete})
+		delete(r.clusters, ctx.namespaceName)
 		return reconcile.Result{}, nil
 	}
+	if err != nil {
+		logger.Errorf("Could not fetch Cassandra definition. Will retry. Error: %v", err)
```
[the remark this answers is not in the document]
sebbonnet: Yes - you're right, at the moment we get duplicate errors in the logs. I'll change that

cassandra-operator/cmd/operator/controller_test.go
```diff
@@ -0,0 +1,587 @@
...
+	AfterEach(func() {
+		fakes.assertAll(GinkgoT())
+	})
+
+	Describe("event dispatching", func() {
+
+		Context("cluster changes", func() {
+			It("should dispatch a DeleteCluster event when no Cassandra definition found", func() {
+				fakes.cassandraIsNotFoundIn(clusterNamespaceName)
+				fakes.dispatcher.On("Dispatch", &dispatcher.Event{
```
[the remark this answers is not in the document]
sebbonnet: sounds good

cassandra-operator/cmd/operator/controller_test.go
```diff
@@ -0,0 +1,587 @@
...
+			BeforeEach(func() {
+				configMap = &corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      desiredCassandra.CustomConfigMapName(),
+						Namespace: desiredCassandra.Namespace,
+					},
+				}
+			})
+
+			It("should dispatch configMap events before Cassandra ones", func() {
```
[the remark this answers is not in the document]
sebbonnet: it doesn't. there are no built-in way in testify and no appetite for either: https://github.com/stretchr/testify/issues/684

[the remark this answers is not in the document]
sebbonnet: we will need to build something specifically for this test to assert the order.

cassandra-operator/cmd/operator/main.go
```diff
@@ -142,17 +144,10 @@ func startOperator(_ *cobra.Command, _ []string) error {
...
-			eventDispatcher:    dispatcher.New(receiver.Receive, stopCh),
-			eventRecorder:      eventRecorder,
-			previousCassandras: map[string]*v1alpha1.Cassandra{},
-			previousConfigMaps: map[string]*corev1.ConfigMap{},
-		},
+		Reconciler: NewReconciler(clusters, mgr.GetClient(), eventRecorder, dispatcher.New(receiver.Receive, stopCh)),
 	})
 	if err != nil {
-		entryLog.Error(err, "unable to set up individual controller")
+		entryLog.Error(err, "unable to set up Cassandra reconciler controller")
```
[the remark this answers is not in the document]
sebbonnet: nope, I'll use Fatal

cassandra-operator/test/e2e/pod_events.go
sebbonnet: The order of the pods doesn't matter.
Whichever pod is deleted first has to be started before the other pod is deleted to ensure there is no downtime

cassandra-operator/test/e2e/pod_events.go
sebbonnet: The basic idea is to validate no 2 pods were down at the same time.
It doesn't matter which pod was first deleted.
If a pod was deleted before the other pod, it should have been (re)created before the other pod is deleted. If that's not the case then 2 pods were down at one moment in time, which is the error case.

alangibson01: OK, that makes sense now. Thanks! Maybe we could rename the function to `PodsNotDownAtTheSameTime` ...?

cassandra-operator/pkg/cluster/state.go
sebbonnet: I've now split state (and its test) into 3 different files that contain both the interface and its implementation:
- state.go exposing the exported interface
- clusterstate.go gathers code related to cluster (statefulset) reconstruction
- snapshot.go gathers code related to snapshot reconstruction

Also factory.go contains the objectRefenceFactory interface and its implementation

cassandra-operator/pkg/cluster/clusterstate.go
```diff
@@ -0,0 +1,239 @@
...
+	sort.SliceStable(statefulSets.Items, func(i, j int) bool {
+		return statefulSets.Items[i].Name < statefulSets.Items[j].Name
+	})
+	return statefulSets, nil
+}
+
+func (cass *currentClusterStateFinder) buildPodSpecFrom(statefulSet *v1beta2.StatefulSet, desiredCassandra *v1alpha1.Cassandra) (v1alpha1.Pod, error) {
+	cassandraContainer, err := cass.containerWithName(cassandraContainerName, statefulSet)
+	if err != nil {
+		return v1alpha1.Pod{}, err
```
alangibson01: Any reason we don't change the return type to `*v1alpha1.Pod, error` and return `nil, err` here?

cassandra-operator/pkg/cluster/state_test.go
```diff
@@ -0,0 +1,483 @@
...
+
+		AfterEach(func() {
+			mockInternalStateFinder.AssertExpectations(GinkgoT())
+		})
+
+		Context("an error occurs when finding the state", func() {
+
+			It("should return an error when unable to find the cluster state", func() {
+				unknownError := fmt.Errorf("unknown")
+				mockInternalStateFinder.On("findClusterStateFor", desiredCassandra).
```
alangibson01: Could we do something similar to what you've done in the other test files, and hide these calls underneath methods that describe the test state that's being set up?

cassandra-operator/pkg/cluster/clusterstate.go
sebbonnet: good idea

cassandra-operator/pkg/cluster/state_test.go
sebbonnet: seems like I got lazy :)
but let me add wrapper methods here too

PR #172, opened 2019-08-21
Expose reconciliation syncPeriod as cmd line arg

sebbonnet:
So users may control the minimum reconciliation frequency.

Default is aggressively set to 5m to workaround the
asynchronous nature of the event processing.

The event processing is done asynchronously to ensure we process updates
for each cluster sequentially; parallel updates to the same cluster
could cause disruption (multiple nodes down) or leave it in an unknown
state. This async processing means that event processing failure cannot be
requeued by returning an error to the reconciler.
To compensate for this, we lower the reconciler "re"sync period, so potential failed
updates are applied within a timely manner.

Required by #53


sebbonnet:
As far as I understand `resync` is used by the informers to resend all events to listeners and does not make calls to the API server. This is not to be confused with `relist` which does call the API server. 
This issue explains it a bit better hopefully: https://github.com/kubernetes/kubernetes/issues/23394

In our case the _listeners_ are the controller that watches for changes in the Cassandra resources and sends reconciliation requests. 
Now I'm not entirely sure what _events_ actually translates too. I would have thought we get 1 event and so 1 reconciliation request for each live Cassandra resource, but it seems we're getting a few more and some for Cassandra definitions that have been long deleted. Regarding the latter I think this could be because there are some dangling resources that are still being watched, as I noticed that we stop seeing reconciliation requests about 5mins after the Cassandra definition is (properly) deleted - which seems to correspond to the watch timeout. But I'm not entirely sure, so will try to find out more.

rsync can be set to as low as `1s`, in fact setting it to anything lower gets changes to the 1s - see https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L393

rsync can be also disabled by setting it to `0`. 
In this case, reconciliation requests will only be sent when a change has been detected for one of the watched resources or once on operator restart. So in other word, no periodic reconciliation. 

On the whole, I think it is best not to enforce a minimum value similarly, but let the underlying informer validate it and change if needs be as done in the shared informer.
So I would actually be leaning towards removing the validation to allow `0` value.

alangibson01:
OK, I'm happy with that, and I think allowing 0 is a good idea too.

sebbonnet:
@alangibson01 I've changed the arg validation to allow the '0' value.

cassandra-operator/cmd/operator/main.go
```diff
@@ -85,8 +85,8 @@ func handleArgs(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("invalid metric-poll-interval, it must be a positive integer")
 	}
 
-	if !isPositive(controllerSyncPeriod) {
-		return fmt.Errorf("invalid controller-sync-period, it must be a positive integer")
+	if !isPositive(controllerSyncPeriod) && controllerSyncPeriod != 0 {
```
alangibson01: Looks a bit clunky. Couldn't we just have an `isNegative` function?

cassandra-operator/cmd/operator/main.go
```diff
@@ -85,6 +85,10 @@ func handleArgs(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("invalid metric-poll-interval, it must be a positive integer")
 	}
 
+	if !isPositive(controllerSyncPeriod) && controllerSyncPeriod != 0 {
+		return fmt.Errorf("invalid controller-sync-period, it must not be a negative integer")
```
alangibson01:
```suggestion
		return fmt.Errorf("invalid controller-sync-period, it must not be a negative duration")
```

Same change required on line 85.

cassandra-operator/cmd/operator/main.go
sebbonnet: yeah - I should have known better :-/

Issue #169, opened 2019-08-15
Migrate to kind to bootstrap test cluster


sebbonnet:
Now that DinD has been retired, we should move onto [kind](https://kind.sigs.k8s.io/) to bootstrap our test cluster.
Retiring in favour of kind: https://github.com/kubernetes-retired/kubeadm-dind-cluster

sebbonnet:
I started to work on that given that my DinD cluster no longer wants to bootstrap on my new workstation due to some iptables issue... so I thought I'd rather make it work with Kind than fighting with iptables for a product that has been retired :) 

Branch: https://github.com/sky-uk/cassandra-operator/tree/kind-cluster

So far I got the build to work against my local KIND cluster, but using a remote docker repository.
Using a local repository is not so obvious judging by this: https://github.com/kubernetes-sigs/kind/issues/602

PR #176, opened 2019-08-28
Switch from dind to kind

sebbonnet:
As dind has been deprecated in favour of kind

Preserves the same features as before, namely:
- create a 4-node cluster
- split the node in 2 availability zones
- create storage class for each AZ
- deploy a local volume provisioner to allow creating local volume
- make the cluster work with a local registry

The default kind context is renamed to "kind" for simplicity.
As kind creates its own kube config file, we use KUBECONFIG env
variable to point to it, to avoid having to specify --kubeconfig arg
for each kubectl commands.

Settting up the local registry was a bit fiddly, as we can no longer
use docker (kind uses containerd) on the nodes to setup the
registry-proxy. Instead we deploy a daemonset on each node to allow
them to pull images from the docker registry on the (local) host.

Required by #169


sebbonnet:
Note: the travis build was failing previously with this error:
```
Unfortunately, an error has occurred:\n\t\t\ttimed out waiting for the condition\n\n\t\tThis error is likely caused by:\n\t\t\t- The kubelet is not running\n\t\t\t- The kubelet is unhealthy due to a misconfiguration of the node in some way (required cgroups disabled)\n\t\t\t- No internet connection is available so the kubelet cannot pull or find the following control plane images:\n\t\t\t\t- k8s.gcr.io/kube-apiserver-amd64:v1.11.10\n\t\t\t\t- k8s.gcr.io/kube-controller-manager-amd64:v1.11.10\n\t\t\t\t- k8s.gcr.io/kube-scheduler-amd64:v1.11.10\n\t\t\t\t- k8s.gcr.io/etcd-amd64:3.2.18\n\t\t\t\t- You can check or miligate this in beforehand with \"kubeadm config images pull\" to make sure the images\n\t\t\t\t  are downloaded locally and cached.\n\n\t\tIf you are on a systemd-powered system, you can try to troubleshoot the error with the following commands:\n\t\t\t- 'systemctl status kubelet'\n\t\t\t- 'journalctl -xeu kubelet'\n\n\t\tAdditionally, a control plane component may have crashed or exited when started by the container runtime.\n\t\tTo troubleshoot, list all containers using your preferred container runtimes CLI, e.g. docker.\n\t\tHere is one example how you may list all Kubernetes containers running in docker:\n\t\t\t- 'docker ps -a | grep kube | grep -v pause'\n\t\t\tOnce you have found the failing container, you can inspect its logs with:\n\t\t\t- 'docker logs CONTAINERID'\ncouldn't initialize a Kubernetes cluster"
 ✗ Starting control-plane 🕹️
time="11:01:49" level=debug msg="Running: /usr/bin/docker [docker ps -q -a --no-trunc --filter label=io.k8s.sigs.kind.cluster --format {{.Names}}\\t{{.Label \"io.k8s.sigs.kind.cluster\"}} --filter label=io.k8s.sigs.kind.cluster=kind]"
time="11:01:49" level=debug msg="Running: /usr/bin/docker [docker rm -f -v kind-worker3 kind-worker4 kind-control-plane kind-worker kind-worker2]"
Error: failed to create cluster: failed to init node with kubeadm: exit status 1
make: *** [kind] Error 1
```

This was fixed by using an officially supported image for kind v0.4.0: `kindest/node:v1.11.10@sha256:176845d919899daef63d0dbd1cf62f79902c38b8d2a86e5fa041e491ab795d33` instead of `kindest/node:v1.11.10`.
Kind v0.5.1 has been realease a few days ago, so I've now used the corresponding node image instead

sebbonnet:
It looks like we gained over 10mins by switching to kind! :smiley: 
Before: https://travis-ci.com/sky-uk/cassandra-operator/builds/124758351
>  Ran for 43 min 19 sec

After: https://travis-ci.com/sky-uk/cassandra-operator/builds/125061587
> Ran for 32 min 29 sec

The cluster setup is about 4mins faster and the operator e2e tests are running faster too (although the snapshot e2e tests are comparable).

test-kubernetes-cluster/recreate-kind-cluster.sh
```diff
@@ -162,76 +153,113 @@ spec:
...
-do
-    numNodes=$((numNodes + numZoneNodes))
-done
-
-DIND_VERSION=${DIND_VERSION:-"1.10"}
-echo "Downloading dind artifacts into ${buildDir}"
-curl -L "https://github.com/kubernetes-sigs/kubeadm-dind-cluster/releases/download/v0.1.0/dind-cluster-v${DIND_VERSION}.sh" -o "${buildDir}/dind-cluster-v${DIND_VERSION}.sh"
-chmod +x ${buildDir}/dind-cluster-v${DIND_VERSION}.sh
+# shellcheck disable=SC2120
+function create_registry_proxy {
```
alangibson01: I think we could do with some commentary around how the proxying is set up, I don't really understand why it's doing what it's doing. It might even be worth breaking this and `run_local_registry` out into a shell script of their own, since it's fairly complex.

alangibson01:
Looks fine, just missing a little bit of doc, I think.

test-kubernetes-cluster/recreate-kind-cluster.sh
sebbonnet: The problem is simple: we need to set the TALK environment variable to the docker host IP address before we invoke the `tcp-proxy` entrypoint. 
I was trying to do that without having to build a custom image with a wrapper script around the `tcp-proxy` entrypoint. 
Will see if I can simplify it a bit

sebbonnet: when I look at it again, it does seem over-complicated :laughing:

Issue #79, opened 2019-05-08
`Pod.Memory` and `Pod.CPU` should be moved to a core.v1.ResourceRequirements field


wallrj:
Part of #33 

Kubernetes already has a `core.v1.ResourceRequirements` type so we can use that in the `CassandraSpec` to define the resources for C* node pods we create (via StatefulSet).

sebbonnet:
Closed by #178

PR #178, opened 2019-09-02
Change CassandraPod spec to use ResourceRequirements

sebbonnet:
This is a breaking API change.
It changes the CassandraPod specs to use `core.v1.ResourceRequirements` Kubernetes type instead of the single value we use for cpu and memory. This effectively allows for both requests and limits to be specified for cpu and memory.
This also includes changes to allow cpu request/limit to be removed - which wasn't possible before.

For #79


sebbonnet:
2 unit tests are failing on travis, but not locally... investigating why:
```
creation of stateful sets
/home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:53
  sidecar container
  /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:288
    should allow CPU bursting configurations [It]
    /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:322
    Expected
        <resource.Quantity>: {
            i: {value: 824634515248, scale: -3},
            d: {Dec: nil},
            s: "100m",
            Format: "DecimalSI",
        }
    to equal
        <resource.Quantity>: {
            i: {value: 100, scale: -3},
            d: {Dec: nil},
            s: "100m",
            Format: "DecimalSI",
        }
    /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:333
```

```
creation of stateful sets
/home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:53
  sidecar container
  /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:288
    should allow Memory bursting configurations [It]
    /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:349
    Expected
        <resource.Quantity>: {
            i: {value: 824634515136, scale: 0},
            d: {Dec: nil},
            s: "50Mi",
            Format: "BinarySI",
        }
    to equal
        <resource.Quantity>: {
            i: {value: 52428800, scale: 0},
            d: {Dec: nil},
            s: "50Mi",
            Format: "BinarySI",
        }
    /home/travis/gopath/src/github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/cluster/cluster_test.go:360
```

sebbonnet:
It looks like the sidecarCpuLimit is not initiliased to the same value all the time when the variable is initialised from an another one:
```
maxSidecarMemoryRequest = resource.MustParse("50Mi")
sidecarMemoryLimit = maxSidecarMemoryRequest
```

Changing the assignment to a new quantity object fixes the problem
```
maxSidecarMemoryRequest = resource.MustParse("50Mi")
sidecarMemoryLimit = resource.MustParse("50Mi")  
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -91,12 +91,22 @@ func validateRacks(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
 
 func validatePodResources(c *v1alpha1.Cassandra, fldPath *field.Path) field.ErrorList {
 	var allErrs field.ErrorList
-	if c.Spec.Pod.Memory.IsZero() {
+	if c.Spec.Pod.Resources.Requests.Memory().IsZero() {
```
alangibson01: Do we need to add additional validation around use cases like limit < request?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -546,6 +574,10 @@ var _ = Describe("ValidateCassandraUpdate", func() {
 	)
 })
 
+func addToQuantity(quantity resource.Quantity) {
```
alangibson01: Unused?

cassandra-operator/pkg/cluster/state_test.go
```diff
@@ -249,8 +249,16 @@ func aClusterDefinition() *v1alpha1.Cassandra {
 			Datacenter: ptr.String("my-datacenter"),
 			Racks:      []v1alpha1.Rack{{Name: "a", Replicas: 1, StorageClass: "rack-storage", Zone: "rack-zone"}},
 			Pod: v1alpha1.Pod{
-				Memory:      resource.MustParse("1Gi"),
-				CPU:         resource.MustParse("100m"),
+				Resources: corev1.ResourceRequirements{
```
alangibson01: Maybe a helper function could be used to make the test more readable.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
sebbonnet: I was in 2 minds about this. On one hand it is duplicating the kubernetes logic which will reject this case, but on the other hand adding the validation is easy and would provide a faster feedback to the users. 
I had a quick look at the implication of not having the validation and the pod never gets created while the operator waits forever for the statefulset to become ready, which is not nice at all. 
As an aside it also shows that our logic in `accessor#WaitUntilRackChangeApplied` is not robust enough to pick up the case where the pod creation has been rejected. We should at least check the statefulset status and return an error if the pod creation failed.
I'll raise an issue for that

sebbonnet: raised this: https://github.com/sky-uk/cassandra-operator/issues/180

cassandra-operator/pkg/cluster/state_test.go
sebbonnet: It would be good to have a ClusteSpecBuilder of some sort, as we repeat ourselves quite a few times in the test. But perhaps do that separately?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra.go
```diff
@@ -111,6 +112,26 @@ func validatePodResources(c *v1alpha1.Cassandra, fldPath *field.Path) field.Erro
 			),
 		)
 	}
+	if isRequestGreaterThanLimit(c.Spec.Pod.Resources.Requests.Memory(), c.Spec.Pod.Resources.Limits.Memory()) {
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				fldPath.Child("Resources.Requests.Memory"),
+				c.Spec.Pod.Resources.Requests.Memory().String(),
+				fmt.Sprintf("must be lower than spec.Pod.Resources.Limits.Memory (%s)", c.Spec.Pod.Resources.Limits.Memory().String()),
```
alangibson01: Technically "must not be greater than" - it's fine if they're equal.

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -54,6 +55,7 @@ var _ = Describe("ValidateCassandra", func() {
 			c := aValidCassandra()
 			mutate(c)
 			err := validation.ValidateCassandra(c).ToAggregate()
+			fmt.Println(err)
```
alangibson01: Is this required?

sebbonnet: damn... thanks for spotting that!

PR #179, opened 2019-09-03
Update issue templates

dudycooly:


.github/ISSUE_TEMPLATE/idea.md
```diff
@@ -1,13 +1,18 @@
...
 about: Suggest an idea for this project
+title: ''
+labels: small-idea
+assignees: ''
 
 ---
 
-### Is your feature request related to a problem? Please describe
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+## NOTE: Please assign this issue to "Planning" project
```
sebbonnet: I don't think we can expect contributors to do that

.github/ISSUE_TEMPLATE/idea.md
```diff
@@ -1,13 +1,18 @@
 ---
-name: Feature request
+name: Idea
 about: Suggest an idea for this project
+title: ''
+labels: small-idea
```
sebbonnet: Why `small-idea`, is there a `big-idea` label?

dudycooly: A big idea which involves multiple deliverables will be 'epic'. Any idea which can be delivered as single deliverable can be qualified as 'small-idea'

.github/ISSUE_TEMPLATE/idea.md
dudycooly: It is best effort to remind them, if not we have to fix that prompty. As we would be moslty operating with a closed user group until we get public traction on this project, I trust this reminder would help us to place the issue under the right project. This is  to avoid growing number of issues missing planning

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
+title: ''
+labels: bug, next-bug
```
alangibson01: I don't think `next-bug` should be in here. This is specific to our way of working internally and how we manage boards and the like, and I don't see its relevance in an open-source project.

.github/ISSUE_TEMPLATE/idea.md
alangibson01: I'm with Seb on this one.

.github/ISSUE_TEMPLATE/bug_report.md
dudycooly: @alangibson01 : next-bug is label used by Wuffle to place bugs under bug column

.github/ISSUE_TEMPLATE/idea.md
dudycooly: The only concern is issues not assigned to Planning project board would miss the planning, and spotting critical ideas outside planning process  would trigger re-prioritisation. As it is just reminder, and won't stop anyone from raising issue without selecting a project, thought it would be useful. I am happy to remove that note if you are going to remember to assign new idea issues to Planning project board without that note,  including ones created by contributors.   Are we happy to take that hit ?

dudycooly: Other idea could be a GitHub bot to assign any new issues reported to Planning project. Unfortunately template wont allow selecting a project

.github/ISSUE_TEMPLATE/bug_report.md
alangibson01: Understood, but Wuffle is only visible internally, not to external contributors. Having an extra label specifically to accommodate the way we use Wuffle doesn't seem right to me.

.github/ISSUE_TEMPLATE/idea.md
alangibson01: OK, fair enough.

.github/ISSUE_TEMPLATE/bug_report.md
dudycooly: I see your point now and agree with you. Yes, it would give an impression that project is being managed outside the community.  " Is it acceptable or not? " is the real question needs answering. 

Because  I see 'next-bug' is no different from  'next' label being used already. As long as we use a common Wuffle to manage  issues across internal and external projects, even if we dont add this label as part of this template. It will be automatically added by Wuffle bot by placing the issue in Bugs column. And Wuffle has much wider footprints than these column labels: Wuffle bots update on labels will be left as issue comments puzzling external contributors

alangibson01: @dudycooly So what's the conclusion? Do we omit the 'next-bug' label? If so, I'm happy with that.

dudycooly: @alangibson01 - As long as we use a common Wuffle, we can't avoid those labels even if I remove it from the template.  We can discuss separately whether to unlink this repo from Wuffle (which is the root cause for your concern: enforcing internal process on this repo) or not?  And about alternatives for a common Wuffle

.github/ISSUE_TEMPLATE/idea.md
sebbonnet: I would remove any mention about our internal process and keep the templates only geared towards external contributions.

.github/ISSUE_TEMPLATE/idea.md
sebbonnet: Again I think we need to gear this towards external contributions and use a more generic label such as `idea` or `feature`. We can re-categorise this internally to `small-idea` or `epic` as we see fit.

dudycooly:
as agreed in standup. merging this

PR #183, opened 2019-09-06
Update issue templates

dudycooly:
As discussed with @sebbonnet ,  renamed as 'small-idea' to 'feature'
and removed note  reminding to select planning as project


sebbonnet:
Follow-up for #179

PR #185, opened 2019-09-12
Removing UseEmptyDir from Cassandra spec

sebbonnet:
This property is no longer needed, as emptyDir can now be defined via
the `rack.storage` - e.g.
```
  racks:
    - name: a
      replicas: 1
      storage:
        - emptyDir: {}
```

Required by #74


sebbonnet:
I usually do the wiki updates when the changes are released, which I'm planning to do once support is added for multiple volumes

PR #188, opened 2019-09-13
Api builders

sebbonnet:
This introduces builders for all the `v1alpha1` types we expose in the API to streamline and make it easier to setup the tests.
Both unit and e2e tests are making use of the api builders except `types_test.go` to avoid circular imports. This is because `types_test.go` which lives in `v1apha1` needs importing builder from `apis` which in turn needs importing types from `v1alpha1`


cassandra-operator/test/e2e/parallel/modification/modification_test.go
```diff
@@ -64,11 +64,11 @@ var _ = Context("Allowable cluster modifications", func() {
 	AfterEach(func() {
 		podWatcher.Stop()
 	})
-
-	AfterEach(func() {
-		DeleteCassandraResourcesForClusters(Namespace, clusterName)
-		resources.ReleaseResource(resourcesToReclaim)
-	})
+	//
```
alangibson01: Do we need this?

sebbonnet: hehe - yes we do :man_facepalming:

PR #190, opened 2019-09-17
Allow multiple rack storages

sebbonnet:
This follows up on #184 to allow creating clusters with multiple custom volumes per rack.
This is so we can support custom volume mounts with different volume types. For instance we envisage that most users would want to have at least 2 volumes to separate logs from data and so have `/var/log/cassandra` and `/var/lib/cassandra` mounted in different volumes, with perhaps the first an `emptyDir` and the second a `PV`. In which case the Cassandra definition would look like this:
```
    racks:
    - name: a
      replicas: 1
      storage:
      - path: "/var/lib/cassandra"
        persistentVolumeClaim:
          resources:
            requests:
              storage: 1Gi
      - path: "/var/log/cassandra"
        emptyDir: {}        
      zone: eu-west-1a
```

When using multiple volumes the `path` must be unique.
It is also used to derive the name of the volume and associated mount in the container, by stripping the lead `/` and replacing remaining `/` with `-`  e.g. a volume mount with path: `/var/lib/cassandra` would have a volume named `var-lib-cassandra`

Custom volumes can be defined when creating the cluster initially, but can also be added at a later stage by editing the Cassandra definition. Volume removal and volume property updates are not supported. 

Required by #78


cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -158,13 +158,12 @@ var _ = Describe("creation of stateful sets", func() {
 				// then
 				volumeClaims := statefulSet.Spec.VolumeClaimTemplates
 				Expect(volumeClaims).To(HaveLen(1))
-				Expect(volumeClaims[0].Name).To(Equal(fmt.Sprintf("cassandra-storage-%d", storageIndex)))
+				Expect(volumeClaims[0].Name).To(Equal("varlibcassandra"))
```
alangibson01: Would it be possible to have this as `var-lib-cassandra`? Just makes it a little more readable.

cassandra-operator/pkg/cluster/clusterstate_test.go
```diff
@@ -150,8 +152,61 @@ var _ = Describe("the cluster state reconstruction", func() {
...
+			Expect(currentCassandra.Spec.Racks[0].Storage[0].PersistentVolumeClaim).To(BeNil())
+			Expect(*currentCassandra.Spec.Racks[0].Storage[0].Path).To(Equal("/emptydir-path-1"))
+			Expect(currentCassandra.Spec.Racks[0].Storage[3].EmptyDir).NotTo(BeNil())
+			Expect(currentCassandra.Spec.Racks[0].Storage[3].PersistentVolumeClaim).To(BeNil())
+			Expect(*currentCassandra.Spec.Racks[0].Storage[3].Path).To(Equal("/emptydir-path-2"))
+		})
+	})
 })
 
+func changeOrderOfVolumeMountsIn(statefulSets *v1beta2.StatefulSetList) {
```
alangibson01: Unused function. Are we missing a test, or is this a leftover we don't need?

cassandra-operator/pkg/cluster/cluster_test.go
sebbonnet: yeah sure - I actually started like that and in fact changed it to remove all special characters in the last commit: 97c0f7bf8dae03fb4474294d97f535dfda9fe3c1
I can simply drop that commit

cassandra-operator/pkg/cluster/clusterstate_test.go
sebbonnet: leftover - I will remove

PR #192, opened 2019-09-18
Allow updates and removal of emptyDir storage

sebbonnet:
As `emptyDir` lifecycle is tied to the pod's and would therefore be recreated from scratch for every `StatefulSet` rolling update.
Keep on preventing updates to persistentVolumeClaim as editing `volumeClaimTemplates` in a `StatefulSet` is not allowed. This may change once support is added for StatetefulSet pvc storage resizing (possibly in 1.17) - see https://github.com/kubernetes/kubernetes/issues/68737

Required by #78


sebbonnet:
I think it would make sense to also include this into the next release and so I will update #191 and WIKI https://github.com/sky-uk/cassandra-operator-wiki/pull/5 accordingly

alangibson01:
I understand the code and I'm happy with it, but I don't understand the motivation - why does having the ability to remove `emptyDir` volumes help? We may need to explain that in the wiki if it's significant.

sebbonnet:
> why does having the ability to remove emptyDir volumes help

If you create it with the wrong path or want to move to a persistent volume.

Issue #161, opened 2019-07-31
Default component versions may be incompatible with the operator


sebbonnet:
When creating a cluster without specifying a version for the sidecar, bootstrapper and snapshot docker images, the `latest` version for each will be used.
The images version may be incompatible with the operator version which can lead to issues that are hard to troubleshoot.

Trying to do some validation in the operator sounds complicated as semantic versioning is not used to denote compatibility between components - e.g. `cassandra-operatror:v1.1.0` may still be backward compatible with `cassandra-bootstrapper:v0.9.0`.
Besides maintaining a compatibility table across all components would be both tedious and hard to test.
 
The safest approach would be to remove usage of `latest` altogether and force users to specify image version they want to use explicitly.

An other alternative would be to default it to the operator version, but this would complicate things when we start applying defaults via a webhook

PR #193, opened 2019-09-18
Default component image version to the operator version

sebbonnet:
Rather than setting it to latest in the `skyuk` repository which may be incompatible with the deployed operator.

When no image names are specified for operator components such as bootstrapper, sidecar and snapshot, the following format is used: `<operatorImageRepositoryPath>/<imageName>:<operatorImageVersion>`.

If  / when using a mutating webhook in the future we could make it use a `noop` image scheme to skip applying default image names and then have the operator apply the defaults based on the operator image name.

Required by #161


sebbonnet:
I initially wasn't sure about this approach, which i found a bit fragile and not compatible with a mutating webhook, but the addition of the `ImageScheme` in the latter commits makes it more appealing as it gives more flexibility and would help dealing with defaults in the mutating webhook. 
Besides this solution is still better than what we have today from a user experience view point.

cassandra-operator/test/e2e/test_environment.go
```diff
@@ -111,9 +112,9 @@ func init() {
 
 	NodeTerminationDuration = NodeStartDuration
 	NodeRestartDuration = NodeStartDuration * 2
-	CassandraBootstrapperImageName = getEnvOrDefault("CASSANDRA_BOOTSTRAPPER_IMAGE", fmt.Sprintf("skyuk/%s:%s", v1alpha1.DefaultCassandraBootstrapperImageName, "latest"))
-	CassandraSidecarImageName = getEnvOrDefault("CASSANDRA_SIDECAR_IMAGE", fmt.Sprintf("skyuk/%s:%s", v1alpha1.DefaultCassandraSidecarImageName, "latest"))
-	CassandraSnapshotImageName = getEnvOrDefault("CASSANDRA_SNAPSHOT_IMAGE", fmt.Sprintf("skyuk/%s:%s", v1alpha1.DefaultCassandraSnapshotImageName, "latest"))
+	CassandraBootstrapperImageName = getEnvOrNil("CASSANDRA_BOOTSTRAPPER_IMAGE")
```
sebbonnet: this is the change that drove all the updates in this commit

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -88,19 +112,24 @@ func setDefaultsForRetentionPolicy(rp *v1alpha1.RetentionPolicy) {
...
-		clusterDefinition.Spec.Pod.BootstrapperImage = ptr.String(v1alpha1.DefaultCassandraBootstrapperImage)
+func setDefaultsForImages(clusterDefinition *v1alpha1.Cassandra, imageScheme ImageScheme) {
+	if imageScheme == nil {
+		imageScheme = &skyUkLatestImageScheme{}
 	}
+
 	if clusterDefinition.Spec.Pod.Image == nil {
 		clusterDefinition.Spec.Pod.Image = ptr.String(v1alpha1.DefaultCassandraImage)
 	}
+	if clusterDefinition.Spec.Pod.BootstrapperImage == nil {
```
sebbonnet: `defaultsName` returns a `*string`, so this check won't change if / when we make the mutation webhook call setDefaults with a noop `ImageScheme`

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -23,14 +23,14 @@ const (
 	// DefaultCassandraImage is the name of the default Docker image used on Cassandra pods
 	DefaultCassandraImage = "cassandra:3.11"
 
-	// DefaultCassandraBootstrapperImage is the name of the Docker image used to prepare the configuration for the Cassandra node before it can be started
-	DefaultCassandraBootstrapperImage = "skyuk/cassandra-bootstrapper:latest"
+	// DefaultCassandraBootstrapperImageName is the name of the Docker image used to prepare the configuration for the Cassandra node before it can be started
```
sebbonnet: renamed from ...Image to ...ImageName to distinguish it from `DefaultCassandraImage` which also contain the version

alangibson01:
@sebbonnet What's an image scheme?

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -12,6 +13,29 @@ import (
 	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/util/ptr"
 )
 
+// ImageScheme is used to determine default names for operator components docker images
+type ImageScheme interface {
+	defaultsName(imageName string) *string
```
alangibson01: Why `defaults` and not `default`?

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -100,76 +100,127 @@ var _ = Describe("Cassandra Helpers", func() {
...
+					var imageDefaultsScheme *TemplatedImageScheme
+
+					BeforeEach(func() {
+						imageDefaultsScheme = &TemplatedImageScheme{
+							ImageVersion:   "v123",
+							RepositoryPath: "dockerRegistry/somepath",
+						}
+					})
+
+					It("should ignore the image schema anb use the 3.11 version of the apache cassandra image", func() {
```
alangibson01:
```suggestion
					It("should ignore the image schema and use the 3.11 version of the apache cassandra image", func() {
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -100,76 +100,127 @@ var _ = Describe("Cassandra Helpers", func() {
...
+						clusterDef.Spec.Pod.Image = nil
+						SetDefaultsForCassandra(clusterDef, imageDefaultsScheme)
+						Expect(*clusterDef.Spec.Pod.Image).To(Equal("cassandra:3.11"))
+					})
+
+					It("should defaults the cassandra bootstrapper image based on the scheme", func() {
```
alangibson01:
```suggestion
					It("should default the cassandra bootstrapper image based on the scheme", func() {
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -100,76 +100,127 @@ var _ = Describe("Cassandra Helpers", func() {
...
+						clusterDef.Spec.Pod.BootstrapperImage = nil
+						SetDefaultsForCassandra(clusterDef, imageDefaultsScheme)
+						Expect(*clusterDef.Spec.Pod.BootstrapperImage).To(Equal("dockerRegistry/somepath/cassandra-bootstrapper:v123"))
+					})
+
+					It("should defaults the cassandra snapshot image based on the scheme", func() {
```
alangibson01:
```suggestion
					It("should default the cassandra snapshot image based on the scheme", func() {
```

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -100,76 +100,127 @@ var _ = Describe("Cassandra Helpers", func() {
...
+						clusterDef.Spec.Snapshot = &v1alpha1.Snapshot{
+							Schedule: "1 23 * * *",
+						}
+						clusterDef.Spec.Snapshot.Image = nil
+						SetDefaultsForCassandra(clusterDef, imageDefaultsScheme)
+						Expect(*clusterDef.Spec.Snapshot.Image).To(Equal("dockerRegistry/somepath/cassandra-snapshot:v123"))
+					})
+
+					It("should defaults the cassandra sidecar image based on the scheme", func() {
```
alangibson01:
```suggestion
					It("should default the cassandra sidecar image based on the scheme", func() {
```

sebbonnet:
That's the name I gave to the interface used to construct the docker image name for the operator components like sidecar and bootstrapper. I'd be glad to hear if you think of a better name :)

cassandra-operator/pkg/cluster/state_test.go
```diff
@@ -255,12 +254,10 @@ func aClusterDefinition() *v1alpha1.Cassandra {
 			WithNoSnapshot().
 			WithRacks(apis.ARack("a", 1).WithDefaults())).
 		Build()
-	v1alpha1helpers.SetDefaultsForCassandra(cassandra)
```
alangibson01: Why isn't this needed any more?

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
sebbonnet: just bad english :laughing:

cassandra-operator/pkg/cluster/state_test.go
sebbonnet: This is no longer needed as we call `WithDefaults()` when using the api builders.

cassandra-operator/test/e2e/test_environment.go
```diff
@@ -87,13 +87,13 @@ func init() {
 		if CassandraImageName == nil {
 			panic("FAKE_CASSANDRA_IMAGE must be supplied")
 		}
-		CassandraInitialDelay = 2
+		CassandraInitialDelay = 3
```
alangibson01: These values are all seconds, I presume?

In that event, I guess there are just some times when our Travis builds run a bit more slowly.

sebbonnet: yes seconds - maybe I'm trying to optmise to much here. Basically adding 1s of delay + 2 more liveness checks of 1s is only giving it 3s more... let's see

Issue #95, opened 2019-05-17
Document CRD fields so they appear in the generated CRD


sebbonnet:
Following the CRD generation via #94, we should add description to all Cassandra properties, so they appear in the generated CRD.

PR #194, opened 2019-09-19
Document CRD fields

sebbonnet:
So they appear in the generated CRD

Required by #95


cassandra-operator/Makefile
```diff
@@ -1,7 +1,6 @@
 .DEFAULT_GOAL := all
 .PHONY: all clean install check build fmt vet lint validate-licences test e2e-te setup go-build go-install docker-image push-image release check-style \
-	validate-crds generate-crds \
-	validate-codegen generate-codegen
+	validate-crds generate-crds validate-codegen generate-codegen generate-all
```
alangibson01: I think you can leave `generate-codegen` out as it's a dependency of `generate-all`

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
```diff
@@ -457,26 +486,37 @@ spec:
...
+                    on each Cassandra node to expose node status.
                   type: string
               required:
               - resources
               type: object
             racks:
               items:
-                description: Rack defines the properties of a rack in the cluster
+                description: Rack defines the rack topology in the cluster, where
+                  a rack has a number of replicas located in the same physical grouping
```
alangibson01:
```suggestion
                  a rack has a number of replicas located in the same failure domain
```

I think this fits better with Kubernetes terminology, although you could argue it's less user-friendly. What do you think?

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
```diff
@@ -457,26 +486,37 @@ spec:
...
                   replicas:
+                    description: The desired number of replicas in the rack.
                     format: int32
                     type: integer
                   storage:
+                    description: The rack storage options.
                     items:
                       description: Storage defines the storage properties shared by
-                        pods in the same rack. Only one type of volume may be specified.
+                        pods in the same rack. Multiple volume types are available
```
alangibson01: This is a bit vague. We should probably state all the volume types that are supported, or link off to somewhere that lists all the ones that are currently supported - even if that's just a comment block somewhere else in the source cod.

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
sebbonnet: Sounds good - I picked it up from the WIKI, so I'll update it there too

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
sebbonnet: doesn't it say that just underneath?
> Multiple volume types are available such as `emptyDir` and `persistentVolumeClaim`.

cassandra-operator/Makefile
sebbonnet: We probably would never have a file called `generate-codegen`. I suppose none of the other targets could be files - except perhaps test, so in reality we should not have to list these words as a PHONY.

cassandra-operator/kubernetes-resources/cassandra-operator-crd.yml
alangibson01: I wish Github allowed you to put a comment against a block, rather than just a single line :)

> The volume types available are expected to be a subset of the volume types defined in k8s.io/api/core/v1/VolumeSource.

That's the sentence I meant to query.

cassandra-operator/Makefile
sebbonnet: I think it probably still a good practice to have all targets listed there - but don't really feel strongly about it.
Are we on the same page on the PHONY ?
https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile#targetText=Generally%20all%20targets%20in%20your,%2C%20distclean%20%2C%20and%20so%20on.

sebbonnet:
@alangibson01 I pushed the description change to the CRD. Can you please re-approve?

PR #205, opened 2019-10-07
Upgrade to java11

sebbonnet:
As new travis builds now appear to be started on the new Xenial environment which uses jdk 11.
This was announced here: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
As this comes with jdk 11, we also have to upgrade gradle and so i took the opportunity to upgrade it to the latest 5.x and dgoss/goss too.


cassandra-bootstrapper/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8-slim
+FROM openjdk:11-slim
```
lw346: Could this use the JRE image (`11-jre-slim`) instead of the JDK image?  69.05MB as opposed to 215.83MB.
Do we care that this build might not be repeatable (i.e. it uses `11-slim` rather than `11.0.4-slim`)?

sebbonnet: Good point, there's no need to use the jdk.
I have different sizes for the images than the one they claimed on docker hub: `11.0.4-jre-slim` at 204MB and `11.0.4-jdk-slim` at 401MB. Not sure why, but jre is still an improvement over jdk
```
$ docker history openjdk:11.0.4-jre-slim
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
2f7748c6f4c6        3 weeks ago         /bin/sh -c #(nop)  CMD ["jshell"]               0B                  
<missing>           3 weeks ago         /bin/sh -c set -eux;   dpkgArch="$(dpkg --pr…   126MB               
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV JAVA_URL_VERSION=11.0…   0B                  
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV JAVA_BASE_URL=https:/…   0B                  
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV JAVA_VERSION=11.0.4      0B                  
<missing>           3 weeks ago         /bin/sh -c { echo '#/bin/sh'; echo 'echo "$J…   27B                 
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV PATH=/usr/local/openj…   0B                  
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV JAVA_HOME=/usr/local/…   0B                  
<missing>           3 weeks ago         /bin/sh -c #(nop)  ENV LANG=C.UTF-8             0B                  
<missing>           3 weeks ago         /bin/sh -c set -eux;  apt-get update;  apt-g…   8.79MB              
<missing>           3 weeks ago         /bin/sh -c #(nop)  CMD ["bash"]                 0B                  
<missing>           3 weeks ago         /bin/sh -c #(nop) ADD file:1901172d265456090…   69.2MB        
```

sebbonnet: This explains the difference in size: https://github.com/docker/hub-feedback/issues/331
> I don't think it'll be possible for Docker Hub to display the uncompressed size (as it never receives the files uncompressed, so wouldn't know what the uncompressed size is). Perhaps the presentation could be improved to make clear that the sizes shown are the compressed size (thus the amount of data that will be sent over the wire)

PR #208, opened 2019-10-14
Run containers as non-root users

### LINKED ISSUE #206: Operator pod should not need to be run as root
At present, in order to run an operator pod, it needs to be run as the root user. This means it will not run successfully on Kubernetes clusters using a pod security policy which prohibits running pods as the root user.

### What would be the expected behavior?
The operator pod should be able to run successfully as a non-root user.

1. Create a deployment of the operator in a Kubernetes cluster which uses a pod security policy that prohibits running as root.
2. Observe that the pod fails to start with a `CreateContainerConfigError` and the following reported event: `Error: container has runAsNonRoot and image will run as root`

### Workaround, if known
If a rolebinding is available within the cluster which allows running as a root user, this can be bound to the service account which runs the operator.

### LINKED ISSUE #207: Cassandra pods should not need to be run as root
At present, in order to for Cassandra pods created by the operator to start successfully, they need to be run as the root user. This means they will not run successfully on Kubernetes clusters using a pod security policy which prohibits running pods as the root user.

### What would be the expected behavior?
Cassandra pods should be able to run successfully as a non-root user.

Provide a link to an example, or an unambiguous set of steps to reproduce this bug. 
Include code to reproduce, if relevant
1. Create a deployment of the operator in a Kubernetes cluster which uses a pod security policy that prohibits running as root, using the workaround in #206 to get the operator to start.
2. Create a Cassandra resource.
3. Observe that the first pod in the cluster fails to start with a CreateContainerConfigError and the following reported event: Error: container has runAsNonRoot and image will run as root

### Workaround, if known
If a rolebinding is available within the cluster which allows running as a root user, this can be bound to the service account which runs the operator.

alangibson01:
Up until now, the operator and all containers started on the provisioned Cassandra pods ran as the root user. This meant that the operator wasn't usable by default within Kubernetes deployments which forbid containers running as the root user.

Part of this restriction was due to use using [phusion/baseimage](https://hub.docker.com/r/phusion/baseimage/) as the base image for the operator and sidecar images, as that requires root-level access to perform certain operations. I've switched these images over to using [alpine](https://hub.docker.com/_/alpine) instead, which solves that problem and also gives us the benefit of a reduced image size.

In order to make these changes testable, I've modified the kind cluster config so that a permissive pod security policy is used in `kube-system` and a restrictive one is used in `test-cassandra-operator`.

Fixes #206 
Fixes #207


cassandra-operator/kubernetes-resources/cassandra-operator-rbac.yml
```diff
@@ -27,7 +27,10 @@ rules:
 - apiGroups: ["core.sky.uk"]
   resources: ["cassandras"]
   verbs: ["list", "get", "watch", "update"]
-
+- apiGroups: ["policy"]
```
sebbonnet: I don't think we should have this here.
We should setup our kind cluster with a ClusterRoleBinding that binds all service accounts to the restricted policy

cassandra-operator/kubernetes-resources/cassandra-snapshot.yml
```diff
@@ -15,6 +15,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods/exec"]
   verbs: ["create"]
+- apiGroups: ["policy"]
```
sebbonnet: same comments as above about the need for a ClusterRoleBinding

cassandra-operator/kubernetes-resources/restricted-pod-security-policy.yml
```diff
@@ -0,0 +1,34 @@
+apiVersion: policy/v1beta1
```
sebbonnet: This belongs to the cluster setup and so in our case should live in the `test-kuberentes-cluster` module

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -132,6 +133,9 @@ func (c *Cluster) CreateStatefulSetForRack(rack *v1alpha1.Rack, customConfigMap
 						c.createInitConfigContainer(),
 						c.createCassandraBootstrapperContainer(rack),
 					},
+					SecurityContext: &v1.PodSecurityContext{
+						RunAsUser: ptr.Int64(999),
```
sebbonnet: perhaps have a constant for that user ID?

cassandra-snapshot/test/e2e/steps.go
```diff
@@ -50,6 +51,9 @@ func createCassandraPod(labels map[string]string, podName string) (*v1.Pod, erro
 			Labels:    labels,
 		},
 		Spec: v1.PodSpec{
+			SecurityContext: &v1.PodSecurityContext{
+				RunAsUser: ptr.Int64(999),
```
sebbonnet: perhaps have a constant for the user ID too

cassandra-operator/hack/deploy.sh
```diff
@@ -30,6 +30,16 @@ function waitForDeployment {
...
 }
 
+function verify_pod_security_policy_restrictions {
+    local context=$1
+    local namespace=$2
+
+    echo "Verifying restricted pod security policy has been applied"
+    kubectl --context ${context} -n ${namespace} delete job policy-test || true
+    kubectl --context ${context} -n ${namespace} apply -f ${resourcesDir}/policy-test-job.yaml
+    ${scriptDir}/../../hack/retry.sh 6 10 ${scriptDir}/expect-container-status.sh ${context} ${namespace} job-name=policy-test CreateContainerConfigError
```
sebbonnet: we should also delete the pod afterwards

cassandra-operator/kubernetes-resources/cassandra-admin-rbac.yml
```diff
@@ -6,7 +6,10 @@ rules:
 - apiGroups: ["core.sky.uk"]
   resources: ["cassandras"]
   verbs: ["*"]
-
+- apiGroups: ["policy"]
```
sebbonnet: we shouldn't need this

cassandra-operator/kubernetes-resources/cassandra-node-rbac.yml
```diff
@@ -6,7 +6,10 @@ rules:
 - apiGroups: ["core.sky.uk"]
   resources: ["cassandras"]
   verbs: ["get"]
-
+- apiGroups: ["policy"]
```
sebbonnet: we shouldn't need this

cassandra-operator/Dockerfile
```diff
@@ -1,5 +1,8 @@
...
+FROM alpine:3.10.2
 
+RUN adduser -S -u 999 cassandra-operator
 COPY build/bin/cassandra-operator /cassandra-operator
 
-ENTRYPOINT ["/sbin/my_init", "--quiet", "--", "/cassandra-operator"]
\ No newline at end of file
+USER 999
+
+ENTRYPOINT ["/cassandra-operator"]
```
sebbonnet: new line missing

test-kubernetes-cluster/registry-proxy.yml
```diff
@@ -0,0 +1,30 @@
...
+      containers:
+        - image: "tecnativa/tcp-proxy"
+          name: tcp-proxy
+          command: ["/bin/sh", "-c"]
+          args:
+            - export TALK=$(/sbin/ip route |  awk '/default/ { print $3 ":5000"}');
+              export LISTEN=:5000;
+              /magic-entrypoint /docker-entrypoint.sh haproxy -f /usr/local/etc/haproxy/haproxy.cfg;
+          ports:
+            - containerPort: 5000
```
sebbonnet: new line missing

PR #210, opened 2019-10-24
Set RunAsGroup and FSGroup in pod security context

### LINKED ISSUE #209: Unable to create /var/lib/cassandra/data on EBS volumes
When provisioning a cluster with EBS as the backing storage, Cassandra pods are unable to start up due to wrong file permissions. Cassandra logs show:
```
WARN  [main] 2019-10-23 16:55:10,326 StartupChecks.java:332 - Directory /var/lib/cassandra/data doesn't exist                                                                                                                                                                                                                                                    
ERROR [main] 2019-10-23 16:55:10,327 CassandraDaemon.java:749 - Has no permission to create directory /var/lib/cassandra/data  
```

This happens since `skyuk/cassandra-operator:v0.76.0-alpha`
I was able to create the same cluster with `skyuk/cassandra-operator:v0.75.0-alpha`.

### Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:
* Kubernetes version: 1.12
* Cassandra operator docker image: `skyuk/cassandra-operator:v0.76.0-alpha`
* Cassandra cluster definition - e.g. `kubectl get cassandra my_cluster -o yaml`: 
```
kind: Cassandra
metadata:
  name: my-jaeger-cassandra
spec:
  datacenter: "jaegerspike"
  pod:
    image: "cassandra:3.11"
    resources:
      limits:
        memory: 2Gi
      requests:
        cpu: "1"
        memory: 2Gi
  racks:
    - name: a
      replicas: 2
      storage:
        - persistentVolumeClaim:
            resources:
              requests:
                storage: 10Gi
      zone: eu-west-1a
    - name: b
      replicas: 2
      storage:
        - persistentVolumeClaim:
            resources:
              requests:
                storage: 10Gi
            storageClassName: standard-zone-b
      zone: eu-west-1b
---
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: standard-zone-a
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
  zone: eu-west-1a
  fsType: xfs
---
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: standard-zone-b
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
  zone: eu-west-1b
  fsType: xfs
```

alangibson01:
Should've done this at the same time as setting RunAsUser, to avoid problems writing to EBS volumes. I've added unit tests to verify that the security context has the appropriate properties set in all of the pods which are created.

Fixes #209


sebbonnet:
The code change looks fine, but I'd like to understand why the integration-test did not pick this up, as we create a PV at the default path `/var/lib/cassandra` similarly to the scenario that failed against an EBS volume.
In fact I was able to reproduce the issue in `kind` when creating a pod with the following:
```
apiVersion: v1
kind: Pod
metadata:
  name: pv-pod
spec:
  securityContext:
    runAsUser: 999
  terminationGracePeriodSeconds: 0
  volumes:
    - name: pv-storage
      persistentVolumeClaim:
        claimName: test-pvc
  containers:
    - name: pv-container
      image: cassandra:3.11
      command:
        - "/bin/sleep"
        - "6000"
      volumeMounts:
        - mountPath: "/var/lib/cassandra"
          name: pv-storage
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: standard-zone-b
```
In this case `/var/lib/cassandra` was mounted as root:root and attempting to start Cassandra then failed to create the data dir as described in the bug.

sebbonnet:
Interestingly when creating a Cassandra pod via the operator, the securityContext has additional group properties:
```
 securityContext:          
    fsGroup: 1              
    runAsUser: 999          
    supplementalGroups:
    - 1
```

group 1 is the daemon user.

When I add the additional property to the simple pod I created in my previous comment, I can then start cassandra.

File system before starting Cassandra
```
cassandra@pv-pod:/$ ls -lta /var/lib/cassandra/                                                                                                                                                                                                                       
total 12                                                                                                                                                                                                                                                              
drwxrwsr-x 2 root daemon 4096 Oct 28 19:51 .                                                                                                                                                                                                                          
drwxr-xr-x 1 root root   4096 Oct 17 02:11 ..     
```

After Cassandra started 
```
cassandra@mycluster-a-0:/$ ls -lta /var/lib/cassandra/
total 28
drwxr-sr-x 7 cassandra daemon 4096 Oct 28 20:03 data
drwxr-sr-x 2 cassandra daemon 4096 Oct 28 20:03 commitlog
drwxrwsr-x 6 root      daemon 4096 Oct 28 20:03 .
drwxr-sr-x 2 cassandra daemon 4096 Oct 28 20:03 hints
drwxr-sr-x 2 cassandra daemon 4096 Oct 28 20:03 saved_caches
drwxr-xr-x 1 root      root   4096 Oct 17 02:11 ..
```

alangibson01:
Hmmm, I'm wondering if it's as simple as a discrepancy between the version of Kubernetes we run in `kind` and the version we run on AWS when doing the tests with EBS. On `kind`, we're still using 1.11.10, which is rather old.

sebbonnet:
After investigating, we found that the reason this bug came up in some circumstances and not in others is because the pod security policies which were running in the two test namespaces we were using were different. In the namespace where the bug did arise, we had the following in the active pod security policy:

```
  fsGroup:
    rule: RunAsAny
```

Which meant that no defaulting was being applied for the `fsGroup` property of the `securityContext` on the Cassandra pods, with the result that the Cassandra process was unable to write to its data volume.

In the namespace where the bug didn't arise, the equivalent setting in the pod security policy was:

```
  fsGroup:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
```

Which meant that the `fsGroup` on the `securityContext` of the Cassandra pods had defaulting applied which explicitly specified the group to use.

The same applies to the `supplementalGroups` definition of the pod security policies in use.

The fix in this PR eliminates this problem by ensuring that `fsGroup` and `runAsGroup` are always specified, so relying defaults supplied by a pod security policy is no longer relevant.

sebbonnet:
@alangibson01 this leaves us with just one thing to do: update the changelog

PR #211, opened 2019-10-29
Upgrade to kubernetes 1.12

sebbonnet:
Upgrade kind to 1.12 and the local volume provisioner to latest compatible version
Upgrade kubectl to 1.13.12  which is the latest minor version compatible with a 1.12 server
Travis runs docker 18.06 which is the latest version compatible with 1.12
Also upgrade controller-tools and controller-runtime to latest; v0.2.2 and v0.3.0 respectively


test-kubernetes-cluster/recreate-kind-cluster.sh
```diff
@@ -10,7 +10,7 @@ function setup_volume_on_node() {
     pv_path="/mnt/pv-zone-${zone}"
     kubectl --context kind label --overwrite node ${node} failure-domain.beta.kubernetes.io/zone=eu-west-1${zone}
     docker exec ${node} mkdir -p /data/vol ${pv_path}/bindmount
-    docker exec ${node} mount -o bind /data/vol ${pv_path}/bindmount
+    docker exec ${node} mount -t ext4 -o bind /data/vol ${pv_path}/bindmount
```
alangibson01: What's the reason for this change?

cassandra-operator/hack/update-crds.sh
```diff
@@ -11,4 +11,5 @@ crd_path="${REPO_ROOT}/kubernetes-resources/cassandra-operator-crd.yml"
 
 output="$(mktemp -d)"
 controller-gen crd:trivialVersions=true paths=./pkg/apis/... output:crd:dir="${output}"
-go run "${REPO_ROOT}/hack/munge-crd.go" "${output}"/core.sky.uk_cassandras.yaml > "${crd_path}"
+#go run "${REPO_ROOT}/hack/munge-crd.go" "${output}"/core.sky.uk_cassandras.yaml > "${crd_path}"
```
alangibson01: Why is the munge file no longer necessary?

test-kubernetes-cluster/recreate-kind-cluster.sh
sebbonnet: This has no effect as we're not mounting a device but merely a directory on an existing filesystem  I think. This is a leftover from a trial - I'll remove it.

cassandra-operator/hack/update-crds.sh
sebbonnet: I think this was needed to generate a CRD compatible with a 1.10 k8 cluster which could support and so required a single `version` property. It was also used to add the CRD scope and setting some additionalProperties used in the schema validation.
After upgrading controller-tools I removed the munge.go and saw that we were only missing the CRD scope (which I fixed using a kubebuilder annotation on the CRD) and an additional header "---". I couldn't find a command line arg or annotation to remove the "header". It's a bit ugly, but doesn't break anything.

test-kubernetes-cluster/recreate-kind-cluster.sh
alangibson01: OK, I'll wait for that change - assign back to me once you're finished.

Issue #202, opened 2019-10-03
Structured logging


sebbonnet:
## Motivation
It would be useful to have structured logging throughout our code base and use a single type of logger to ensure controller-runtime reconciler logs are reported.

## Requirements
Use the `go-logr` package.

---- To be filled in by Assignee -----
## Tasks/Feature Set
List tasks/features in order to meet the requirements listed above

PR #212, opened 2019-11-01
Implement operation interruption

### LINKED ISSUE #180: Updating racks waits indefinitely when pod fails to create
### Is your feature request related to a problem? Please describe
When creating a statefulset we need to cater for scenarios where the pod(s) cannot be created for whatever reasons - e.g. insufficient resources, insufficient privileges.
At the moment the operator waits forever for the rack changes to be applied (see `accessor#WaitUntilRackChangeApplied`). As we only allow 1 queued operation per cluster, this  means that the Cassandra definition cannot be updated to rectify the problem (e.g. lowering resources requirements), leaving the user with no choice but restart the operator in order to apply the new Cassandra definition.

### Describe the solution you'd like
I think we need to distinguish cases where the pod creation fails due to an invalid definition (e.g. memory requests > limits) or due to external factors (insufficient resource, privileges).
- invalid Cassandra definition according to Kubernetes: we should ideally catch that upfront when validating Cassandra updates. However this may not always be possible, but we should at least try to identify the affected scenarios.
- invalid Cassandra definition due to human error: when for instance the wrong storage class was used or too much resource has been requested. In this case we would want to be able to amend the Cassandra definition - this would mean somehow cancelling the current update. Perhaps we could consider cancelling the current update if a new update has arrived or find a more deterministic way (perhaps via annotation?) to inform the operator to cancel the current operation?  
- valid Cassandra definition, but the environment cannot meet the requirements. For instance, they are insufficient resources or privileges in the cluster to accommodate the requirements. In this case, it is probably the right thing to do let the operator wait indefinitely, as we would expect the environment to be changed to satisfy the requirements in the Cassandra definition

### Additional context
This issue was found when trying to create a cluster with invalid pod resources where the memory limit was lower than the requested limit. Even though this is a scenario we should catch upfront when validating the Cassandra specs, there are other similar scenarios where the pod will fail to create.

alangibson01:
Introduced a new controller, the Interrupter. The purpose of this is to spot when a cluster definition has changed, check whether the Reconciler is waiting for a statefulset creation from a previous reconciliation attempt to complete, and if so, interrupt the Reconciler.

This has been implemented through a shared synchronised map of ActiveReconciliation structs. Each of these records the revision number of the Cassandra and ConfigMap being used for a revision, along with a channel which can be used by the Interrupter to communicate with the Reconciler. The map is populated with new ActiveReconciliations by the Reconciler, which closes the
channel and removes the record from the map when the reconciliation completes.

In the Reconciler's statefulset wait loop, it checks on every poll whether any message has been received on the channel, and if so, it exits the polling by returning an ErrReconciliationInterrupted. This is then used by the various operations to signify that the reconciliation attempt should not progress any further.

On the Interrupter side, it checks the Cassandra and ConfigMap versions in the ActiveReconciliation, and if either of these differs from the current versions fetched from the API server, it will send a message down the channel to effect the interruption.

Fixes #180


cassandra-operator/pkg/operator/event/receiver.go
```diff
@@ -53,16 +53,14 @@ func (r *OperatorEventReceiver) Receive(event *Event) error {
...
 
-	result := &multierror.Error{}
-	for _, operation := range operationsToRun {
-		logger.Debugf("Executing operation %s", operation.String())
-		if err := operation.Execute(); err != nil {
-			result = multierror.Append(result, fmt.Errorf("error while execution operation %s: %v", operation.String(), err))
-		}
+	stopped, err := r.operationComposer.Execute(operationsToRun)
+	if stopped {
+		logger.Info("Event complete - not all operations were executed")
```
sebbonnet: I would log the operation stopped in the operation composer at info level , so we can see which operation was interrupted. If you can to continue logging the event was stopped here than do it at debug level

cassandra-operator/pkg/operator/reconciler.go
```diff
@@ -259,12 +281,18 @@ func (r *CassandraReconciler) determineChangesForCassandraService(ctx *requestCo
 }
 
 func newContext(request reconcile.Request) *requestContext {
+	logger := "reconciler.go"
```
sebbonnet: not convinced this is necessary, but is sort of hard to read.
maybe wrap this in a method at least?

sebbonnet: if we really want this, then we should apply that change across the board.
I feel it may be best to wait for #202?

cassandra-operator/test/e2e/cluster_setup.go
```diff
@@ -109,7 +109,7 @@ func (c *ClusterBuilder) IsDefined() {
 	}
 
 	if !c.withoutCustomConfig {
-		_, err := customCassandraConfigMap(Namespace, c.clusterName, c.extraConfigFile)
+		_, err := customCassandraConfigMap(Namespace, c.clusterName, false, c.extraConfigFile)
```
sebbonnet: Can you explain why this was necessary?

cassandra-operator/test/e2e/parallel/modification/interrupted_operation_test.go
```diff
@@ -0,0 +1,168 @@
...
+			// then
+			Eventually(StatefulSetsForCluster(Namespace, clusterName), 2*NodeStartDuration, 2*time.Second).Should(HaveLen(2))
+			Expect(ClusterConfigHashForRack(Namespace, clusterName, "b")).ToNot(Equal(ClusterConfigHashForRack(Namespace, clusterName, "a")))
+		})
+	})
+})
+
+func aStuckClusterExists(clusterName string, withCustomConfig bool) {
+	registerResourcesUsed(2)
+	racks := []v1alpha1.Rack{Rack("a", 1), Rack("b", 1)}
```
sebbonnet: In the interest of keeping the test duration short, could we do with a single node?

cassandra-operator/pkg/operator/operator.go
```diff
@@ -69,21 +72,27 @@ func New(kubeClientset *kubernetes.Clientset, cassandraClientset *versioned.Clie
...
 		MaxConcurrentReconciles: 100,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("unable to set up Cassandra reconciler controller: %v", err)
+		return nil, fmt.Errorf("unable to set up Cassandra reconciler reconcileController: %v", err)
 	}
 
+	interrupter, err := controller.New("interrupter", mgr, controller.Options{
+		Reconciler:              NewInterrupter(&activeReconciliations, mgr.GetClient()),
+		MaxConcurrentReconciles: 100,
```
sebbonnet: perhaps make this a constant

cassandra-operator/pkg/operator/controller_test.go
```diff
@@ -512,21 +517,10 @@ var _ = Describe("reconciliation", func() {
 				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("failed to lookup Cassandra service: %s", unknownError)))
 			})
 
-			It("should re-enqueue a single request when multiple error occur", func() {
```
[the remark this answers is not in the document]
alangibson01: Because the ConfigMap is now fetched before we start any reconciliation, it's not possible for multiple errors to occur in the way that this test did, which is why I rather lazily removed this test.

But you're right, it should still be possible to mock out the reconciliation in such a way that we get multiple errors back.

cassandra-operator/pkg/operator/reconciler.go
alangibson01: I think we should wrap this in a method, to make the intent clear. It's helpful for this case, as we have two controllers which want to log messages and it's important to be able to see which controller is logging what message. I don't think we should be waiting until we do #202.

cassandra-operator/test/e2e/cluster_setup.go
alangibson01: I added a new test helper method `TheCustomConfigIsModifiedForCluster` which requires all the same logic around populating the data in the ConfigMap. The only part that's different is whether the `Get` or `Update` API is used to communicate with Kubernetes.

cassandra-operator/test/e2e/parallel/modification/interrupted_operation_test.go
alangibson01: Unfortunately not. Partly because of https://github.com/kubernetes/kubernetes/issues/67250, the only way to properly detect that a change has been successfully made to a stuck cluster is to have a second rack/statefulset against which to verify the change because the first rack will remain stuck until the stuck pods are manually deleted.

sebbonnet: I'm not sure I follow as that bug means that pod spec will not get updated unless the pod is deleted after the statefulset spec was changed.
As long as the tests assert statefulset specs (which they do), they will find that it has changed when the cluster update was processed.

cassandra-operator/test/e2e/cluster_setup.go
sebbonnet: ok

cassandra-operator/test/e2e/parallel/modification/interrupted_operation_test.go
alangibson01: I wanted there to be tests which showed not just that the statefulset spec was updated, but also that pods created from those specs would start successfully.

cassandra-operator/test/e2e/parallel/modification/interrupted_operation_test.go
```diff
@@ -0,0 +1,168 @@
...
+		resources.ReleaseResource(resourcesToReclaim)
+	})
+
+	Context("cluster definition changes", func() {
+		It("should abandon a stuck operation when the cluster definition is updated", func() {
+			// given
+			aStuckClusterExists(clusterName, withoutCustomConfig)
+
+			// when
+			TheBootstrapperImageImmutablePropertyIsChangedTo(Namespace, clusterName, *CassandraBootstrapperImageName)
```
sebbonnet: `CassandraBootstrapperImageName` can be null.
I tried running this test in IntelliJ (with no environment variable) and it paniced due to the nil image. 
All tests should be runnable via IntelliJ without having to add custom env variable.
This should be easily fixed as the image name is not important here as long as it exists.

cassandra-operator/test/e2e/parallel/modification/interrupted_operation_test.go
sebbonnet: I think there is no need - that is not the point of this feature.
I can't see where your tests were checking that the pods started successfully, so the assertions can broadly remain the same.
You could also check an event is generated for additional cover - the additional event is mentioned here: https://github.com/sky-uk/cassandra-operator/pull/212#discussion_r342028376

alangibson01:
@sebbonnet I've made some further changes to the way active reconciliation records are used in communication between the interrupter and the reconciler. The previous way occasionally resulted in race conditions where the interrupter would try to send a message down a closed channel if an reconciliation completed around the same time the interruption was made.

The new approach uses methods on the active reconciliation record itself to regulate access to the communication channel, along with checks which ensure that there'll never be an attempt to close a channel twice, or send a message down a closed channel. All this is wrapped with mutex locks to keep things safe.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+## 0.78.0-alpha
+
+- [BUGFIX] [Implement operation interruption](#212) (@alangibson01)
```
sebbonnet: I would have made this a feature :)

alangibson01:
@sebbonnet I think I've resolved the concurrency issues and race conditions with the last commit in this PR. I think this is ready for another review now.

cassandra-operator/pkg/operator/interrupter.go
```diff
@@ -0,0 +1,88 @@
...
+		return retryReconciliation("unable to fetch Cassandra definition", err)
+	}
+
+	targetConfigMap := i.objectFactory.newConfigMap()
+	err = i.client.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: fmt.Sprintf("%s-config", request.Name)}, targetConfigMap)
+	if unexpectedError(err) {
+		return retryReconciliation("unable to fetch Cassandra custom config definition", err)
+	}
+
+	if activeReconciliation, ok := i.activeReconciliations.Load(clusterName); ok {
```
sebbonnet: How does the interrupter know at this stage that the reconciler would had had the time to start processing the reconcile request? If we have 2 cluster operations in quick successions (e.g. update 1, and update 2), it seems possible that the interrupter could process the 2 requests before the reconciler had the chance to "start" the reconciliation.

cassandra-operator/pkg/operator/interrupter.go
```diff
@@ -0,0 +1,88 @@
...
+func NewInterrupter(activeReconciliations *sync.Map, client client.Client, recorder record.EventRecorder) *Interrupter {
+	return &Interrupter{
+		activeReconciliations: activeReconciliations,
+		client:                client,
+		objectFactory:         &defaultReferenceFactory{},
+		eventRecorder:         recorder,
+	}
+}
+
+var interruptMessage = struct{}{}
```
sebbonnet: this isn't used

cassandra-operator/pkg/operator/reconciler.go
```diff
@@ -41,28 +44,30 @@ type requestContext struct {
...
-		stateFinder:    cluster.NewStateFinder(client),
-		operatorConfig: operatorConfig,
+		clusters:              clusters,
+		client:                client,
+		eventRecorder:         eventRecorder,
+		eventReceiver:         eventReceiver,
+		objectFactory:         &defaultReferenceFactory{},
+		stateFinder:           cluster.NewStateFinder(client),
+		operatorConfig:        operatorConfig,
+		activeReconciliations: activeReconciliations,
```
sebbonnet: `activeReconciliations` is not used

cassandra-operator/pkg/operator/interrupter.go
alangibson01: I'm not convinced this is a problem because the reconciliation process is level-based not edge-based - the Reconciler always tries to bring the cluster into line with whatever the latest version of the Cassandra resource is.

So in the use case you mention:

0. Initial state - no updates in process for cluster X.
1. Edit E1 made to cluster X, reconcile request R1 triggered.
2. Edit E2 made to cluster X, reconcile request R2 triggered.
3. Interrupter picks up R1, finds nothing in `activeReconciliations`, does nothing.
4. Interrupter picks up R2, finds nothing in `activeReconciliations`, does nothing.
5. Reconciler picks up R1, reconciles cluster to latest state defined following E2.
6. Reconciler picks up R2, finds it's already in the state defined following E2, does nothing.

alangibson01:
> With the latest change not every reconciliation can be interrupted, as active reconciliations are only created during a cluster operation that may take a long time - was this done to explicitly prevent short-lived operations to be interrupted?

Yes, I didn't see any need to interrupt operations that are short-lived.

> How do we guard against the interrupter from processing reconcile requests that have not had the chance to be processed by the reconciler (and so may not have reached the wait loop)?

I've addressed this in more detail in my reply to your inline comment. The short summary is that I don't think this is really an issue.

cassandra-operator/pkg/operator/reconciler.go
```diff
@@ -41,13 +44,14 @@ type requestContext struct {
...
-	stateFinder    cluster.StateFinder
-	operatorConfig *Config
+	clusters              map[types.NamespacedName]*v1alpha1.Cassandra
+	client                client.Client
+	eventRecorder         record.EventRecorder
+	eventReceiver         event.Receiver
+	objectFactory         objectReferenceFactory
+	stateFinder           cluster.StateFinder
+	operatorConfig        *Config
+	activeReconciliations *sync.Map
```
sebbonnet: `activeReconciliations` is not used

Issue #213, opened 2019-11-26
initContainers should not request CPU


sebbonnet:
### What is the bug being reported?
The definition of the statefulset for a Cassandra rack requests the same amount of CPU for each initContainer as is requested for the main Cassandra container. This can cause scheduling problems on Kubernetes clusters which use the CPU Management feature.

### What would be the expected behavior?
initContainers should not request any CPU.

### If repeatable, how can we reproduce this issue?
Create any Cassandra cluster with the operator and look at the initContainer definitions on the statefulsets. You'll see that the initContainers are requesting the same amount of CPU as is requested for the main container in the Cassandra spec.
This may result in the init container start up to fail as it attemps to request additional CPU on top of the ones already allocated to the main cassandra container - e.g. `Internal PreStartContainer hook failed: not enough cpus available to satisfy request`

### Workaround, if known
Scale down the operator and manually modify the statefulsets.
 
### Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:
* Kubernetes version: 1.12
* Cassandra operator docker image: v0.77.0-alpha
* Cassandra cluster definition - e.g. `kubectl get cassandra my_cluster -o yaml`: 
```yaml
apiVersion: core.sky.uk/v1alpha1
kind: Cassandra
metadata:                                              
  name: my-name
  namespace: my-namespace
spec:
  datacenter: test
  pod:
    image: cassandra:3.11
    resources:
      limits:
        cpu: "4"
        memory: 2Gi
      requests:
        cpu: "4"
        memory: 2Gi
  racks:
  - name: a
    replicas: 1
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard-zone-a
    zone: eu-west-1a
  - name: b
    replicas: 1
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard-zone-b
    zone: eu-west-1b

```
* Statefulset definition:
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  [ ... ]
spec:
  podManagementPolicy: OrderedReady
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-cassandra
      rack: a
      sky.uk/cassandra-operator: my-cassandra
  serviceName: my-cassandra
  template:
    metadata:
      annotations:
[...]

PR #215, opened 2019-12-02
Assign no cpu & min mem to init containers

sebbonnet:
As init containers are shortlived and lightweight containers
and so should have very little resource requirements.
A minimum memory request is assigned provided that it stays within the
limit of the pod memory requests.

For #213


alangibson01:
@sebbonnet Why aren't the initContainers given a memory limit that matches the requests?

sebbonnet:
@alangibson01 that was a mistake - I have now sets limits to match requests

PR #221, opened 2019-12-12
Use standard labels on generated resources

### LINKED ISSUE #214: Node anti-affinity rules do not account for namespaces
If there are two Cassandra definitions with the same name but in different namespaces, the node anti-affinity rules defined on the statefulsets created for those definitions will prevent Cassandra pods from the two clusters being scheduled to the same node. This is an unnecessary restriction.

### What would be the expected behavior?
Cassandra pods from two independent clusters should be schedulable on the same node, regardless of what their name is or what namespace they are defined in.

Apply the following YAML:
```yaml
---
kind: Cassandra
metadata:
  name: my-cassandra
  namespace: ns1
spec:
  datacenter: dc1
  pod:
    image: cassandra:3.11
    resources:
      limits:
        memory: 2Gi
      requests:
        cpu: "4"
        memory: 2Gi
  racks:
  - name: a
    replicas: 2
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
    zone: eu-west-1a
  - name: b
    replicas: 2
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard-zone-b
    zone: eu-west-1b
---
kind: Cassandra
metadata:
  name: my-cassandra
  namespace: ns2
spec:
  datacenter: dc1
  pod:
    image: cassandra:3.11
    resources:
      limits:
        memory: 2Gi
      requests:
        cpu: "4"
        memory: 2Gi
  racks:
  - name: a
    replicas: 2
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
    zone: eu-west-1a
  - name: b
    replicas: 2
    storage:
    - persistentVolumeClaim:
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard-zone-b
    zone: eu-west-1b
```

When these clusters are provisioned by the operator, Cassandra pods from `ns1/my-cassandra` and `ns2/my-cassandra` will never be scheduled to the same Kubernetes node.

### Workaround, if known
Ensure that Cassandra definitions in different namespaces have different names.
 
### Additional context
This problem would largely only affect Cassandra deployments where the number of pods per rack is close to the number of nodes in the Kubernetes cluster.

### Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:
* Kubernetes version: 1.12
* Cassandra operator docker image: v0.78.0-alpha

alangibson01:
The following standard labels will be added to all resources created by the operator:

- app.kubernetes.io/name
  => non-namespaced name of the Cassandra resource which it relates to
- app.kubernetes.io/instance
  => namespaced name of the Cassandra resource which it relates to
- app.kubernetes.io/version
  => version of the "main" Cassandra image used in the cluster
- app.kubernetes.io/managed-by
  => always set to "cassandra-operator"
- app.kubernetes.io/component
  => for Cassandra pods, not present
  => for snapshot pods, set to "snapshot"
  => for snapshot cleanup pods, set to "snapshot-cleanup"

The "app" label has been removed. Because this was used rather extensively to filter resources down to those created for a particular Cassandra cluster, there have been numerous changes to the label expressions used.

With these labels in place, we then use app.kubernetes.io/instance in the pod anti-affinity selector, as this is unique across all Cassandra resources within the cluster.

Fixes #214


cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -115,6 +130,21 @@ func (c *Cluster) QualifiedName() string {
...
 }
 
+// ApplicationVersion returns the version part of the main Cassandra image name
+func (c *Cluster) ApplicationVersion() string {
+	mainImage := c.definition.Spec.Pod.Image
+	if mainImage == nil {
+		return ""
+	}
+
+	versionSeparatorIndex := strings.LastIndex(*mainImage, ":")
```
sebbonnet: We also have a `extractImageVersion` (and `extractRegistryPath`) in `main.go`.
Could we move this to the `util` package?
Your function seems to do a better job than ``extractImageVersion` which i think would not work when no version is specified :-/ - so reusing your function will fix that bug :)

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -127,19 +157,21 @@ func (c *Cluster) CreateStatefulSetForRack(rack *v1alpha1.Rack, customConfigMap
 		Spec: appsv1.StatefulSetSpec{
 			Selector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					OperatorLabel: c.definition.Name,
-					RackLabel:     rack.Name,
-					"app":         c.definition.Name,
+					ApplicationInstanceLabel: c.QualifiedName(),
+					ManagedByLabel:           ManagedByCassandraOperator,
+					RackLabel:                rack.Name,
```
sebbonnet: The rack label looks a bit out of place compared with the rest :laughing: 
But I think that's ok  asusing `app.kubernetes.io/part-of` will not tell you that it's part of a _rack_ unless the rack name is prefixed with `rack` or similar
```
    app.kubernetes.io/instance: test-cassandra-operator.mycluster-cg8jw
    app.kubernetes.io/managed-by: cassandra-operator
    app.kubernetes.io/name: mycluster-cg8jw
    app.kubernetes.io/version: "3.11"
    rack: a
```

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -214,7 +247,8 @@ func (c *Cluster) CreateService() *v1.Service {
 		ObjectMeta: c.objectMetadataWithOwner(c.definition.Name, "app", c.definition.Name),
```
sebbonnet: I don't think we need the "app" label anymore here
This is a dupe of `app.kubernetes.io/name`

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -282,6 +327,7 @@ func (c *Cluster) CreateSnapshotCleanupJob() *v1beta1.CronJob {
 
 	return c.createCronJob(
 		c.definition.SnapshotCleanupJobName(),
+		"snapshot-cleanup",
```
sebbonnet: This should use the constant you defined in cluster.SnapshotCleanupCronJob

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -337,7 +383,19 @@ func (c *Cluster) createCronJob(objectName, serviceAccountName, schedule string,
...
+	labels := map[string]string{
+		ApplicationNameLabel:     c.Name(),
+		ApplicationInstanceLabel: c.QualifiedName(),
+		ManagedByLabel:           ManagedByCassandraOperator,
+	}
+
+	imageName := c.definition.Spec.Pod.Image
+	if imageName != nil {
+		i := *imageName
+		imageVersion := i[strings.LastIndex(i, ":")+1:]
```
sebbonnet: why not using `ApplicationVersion()`?

cassandra-operator/pkg/cluster/accessor.go
```diff
@@ -88,6 +88,12 @@ type Accessor interface {
 	CreateStatefulSetForRack(c *Cluster, rack *v1alpha1.Rack, customConfigMap *v1.ConfigMap) (*v1beta2.StatefulSet, error)
 }
 
+// SnapshotCronJob is the value of app.kubernetes.io/component for snapshot cronjobs
+const SnapshotCronJob = "snapshot"
```
sebbonnet: Could we move these to cluster.go alongside the other labels?

cassandra-operator/pkg/cluster/cluster.go
alangibson01: I think we could probably do with namespace-qualifying it. Maybe `sky.uk/cassandra-operator/rack`?

cassandra-operator/pkg/cluster/cluster.go
alangibson01: No reason, I'll change it.

alangibson01:
> Could we also update the labels to the operator itself as part of this?

Do you mean the labels on the operator and pod deployment? If so, I’m not sure that’s really for us to worry about too much. We can make recommendations on how people label their deployments, but ultimately it’s up to them.

alangibson01:
@sebbonnet Back to you!

sebbonnet:
@alangibson01 once this is merged, will you update the metrics section in the WIKI?

alangibson01:
@sebbonnet Yes, I'll update the wiki.

PR #230, opened 2020-10-23
Remove app label references

chuckydev:
Required by https://github.com/sky-uk/cassandra-operator/issues/231

Originally the operator was using the `app` label in In order to obtain the IP of each cassandra pod. Thas has since been removed and has caused a bug which defaults the lookup to only fetch from a single pod.

A simple fix for this is to instead lookup using a different label which will always be present. This PR opts for the Application Instance label that defaults to the cluster name.

I have also added a unit test that explicitly checks for the specific label in the podGetter stub.

### DIFF
--- a/cassandra-operator/pkg/metrics/metrics_test.go
+++ b/cassandra-operator/pkg/metrics/metrics_test.go
@@ -2,11 +2,12 @@ package metrics
 
 import (
 	"fmt"
-	"github.com/sky-uk/cassandra-operator/cassandra-operator/test/apis"
 	"math/rand"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/sky-uk/cassandra-operator/cassandra-operator/test/apis"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


cassandra-operator/pkg/metrics/metrics_test.go
```diff
@@ -161,14 +162,17 @@ var _ = Describe("Cluster Metrics", func() {
 
 var _ = Describe("Metrics URL randomisation", func() {
 	var cluster *cluster.Cluster
+	var standardPodLabels map[string]string
 
 	BeforeEach(func() {
 		cluster = aCluster("testcluster", "test")
+		standardPodLabels = make(map[string]string)
```
sebbonnet, line 169: you could initialise the map in one line: `standardPodLabels = map[string]string{"app.kubernetes.ivo/instance": "testcluster"}`

cassandra-operator/pkg/metrics/prometheus.go
```diff
@@ -212,7 +212,7 @@ func (u *randomisingJolokiaURLProvider) URLFor(cluster *cluster.Cluster) string
 }
 
 func (u *randomisingJolokiaURLProvider) podsWithIPAddresses(cluster *cluster.Cluster) ([]v1.Pod, error) {
-	podList, err := u.podsGetter.Pods(cluster.Namespace()).List(metaV1.ListOptions{LabelSelector: fmt.Sprintf("app=%s", cluster.Name())})
+	podList, err := u.podsGetter.Pods(cluster.Namespace()).List(metaV1.ListOptions{LabelSelector: fmt.Sprintf("app.kubernetes.io/instance=%s", cluster.Name())})
```
sebbonnet, line 215: You should use ` cluster.ApplicationInstanceLabel` constant instead of hardcoding the label name.
`cluster *cluster.Cluster` function arg is clashing with the `cluster` package name, so you'll need to rename the parameter so you can reference the constant. Just a small tip as that might not be obvious :)

sebbonnet (commented):
Thanks for your first contribution @chuckydev :). 
I have a few minor comments.

We should release that fix too, which would also give you the chance to see the release process. But we'll need an entry in the CHANGELOG.md first

sebbonnet (approved):

Issue #231, opened 2020-10-23
Metrics always reading from the same node


chuckydev:
### What is the bug being reported?

A log was observed when running the operator, which implied the metrics were being fetched from the same node. This implies that the IP lookup is not working as expected and we are resorting to fetching metrics from the same node.
```
time="2020-10-21T19:44:05Z" level=info msg="No pods with IP addresses found for cluster ccl-prod.ccl-cassandra. Falling back to the cluster service name for jolokia url."
```
The original issue is #3. This appears to have occured since the removal of the app label from each cassandra pod.

### What would be the expected behavior?
The IP Lookup to search for the correct label and return a list of IPs. This then allows the metrics to be fetched from a random node.

### If repeatable, how can we reproduce this issue?
I believe this should exist in versions since the removal of the app label

adamdougal:
Need to chat to @chuckydev to get more of an understanding of what this is about.

chuckydev:
This can be closed as the tag we look up on each operator pod has been corrected

PR #232, opened 2020-10-27
Used the Qualified Name for looking up jolokia metrics

chuckydev:
In order to look up jolokia names using pod IPs I raised #230 . When trying to run this it seems like the issue #231 still exists. This is due to the fact that the instance label on pods is set to the "Qualified" name not the standard cluster name.

This PR should correct the contents of the label in test and lookup based on the correct QualifiedName instead.


sebbonnet (commented):
lgtm. can you update the CHANGELOG accordingly too?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+
+## 0.83.0-alpha
```
sebbonnet, line 4: We normally increase the patch section for bug fix, so the tag could be `0.82.1-alpha` the first time and so this one `0.82.2-alpha`. But I'm happy to do this next time.

sebbonnet (approved):
We normally increase the patch section for bug fix, so the tag could be 0.82.1-alpha the first time and so this one 0.82.2-alpha. But I'm happy to do this next time.

sebbonnet (approved):

Issue #229, opened 2020-10-21
Cassandra Resource Validation Metrics


chuckydev:
## Motivation
When an invalid change is made to the cassandra resource there is currently no visibility as to whether the change has successfully passed validation or not. 

This can be alerted on in the case that someone was expecting a change to their cluster but it was not applied. I.e. If a PVC increase was attempted but it is not currently supported so it fails validation.

## Requirements
- Per Cluster metric stating whether a resource has succeeded or failed.

### Additional context
- The reconciler metrics are currently not sufficient as the validation simply logs an error to events and passes nil to the reconciler. Otherwise we could have used the reconciler metrics on port 8080. If the reconciler fails we have the `controller_runtime_reconcile_errors_total` which is not cluster specific.

---- To be filled in by Assignee -----
## Tasks/Feature Set
List tasks/features in order to meet the requirements listed above

Bez625:
Discussion is here https://sky.slack.com/archives/CB41RC81H/p1603199364047900

Bez625:
When a Cassandra Resource is validated the spec is examined and validations errors are collected. If one or more validation errors are detected we record an InvalidChange event with the "Message" being aggregated errors as a string.

In order to provide a metric around this - considering the reported case of a PVC change not being supported - we have a few options:

#### 1. We simply increase a failed validation count for the specific Cassandra cluster and namespace:
```
# HELP cassandra_failed_validation_count Total number of failed validations in a Cassandra cluster
# TYPE cassandra_failed_validation_count count
cassandra_failed_validation_count{cluster="some-cassandra",namespace="some-namespace"} 1
```
**PROS:** This is a simple change and covers all validation failures.
**CONS:** It does not help identify why the validation failed. That would require analysis from the user.

#### 2. We only increase metrics for validation failures we are specifically prepared to handle:
```
# HELP cassandra_failed_validation_count Total number of failed validations in a Cassandra cluster
# TYPE cassandra_failed_validation_count count
cassandra_failed_validation_count{cluster="some-cassandra",namespace="some-namespace",reason="PVC_update"} 1
```
**PROS:** The metrics are more informative and allow for the ingester to have better control over their use.
**CONS:** Requires specific development each time a new reason is handled. If there are multiple failures we have to increment multiple metrics and this could lead to multiple alerts for the same validation failure. This does not report unhandled validation failures.

#### 3. Increase metrics for specific handled validation failures and have a catch all for unhandled validation failures.
```
[...]

PR #233, opened 2020-10-30
Add a Failure Metric Increment Event based on failed CRD validation

Bez625:
When an invalid change is made to the cassandra resource there is currently no visibility as to whether the change has successfully passed validation or not. 

This PR adds an IncrementValidationFailures method to the prometheus metrics `cassandra_failed_validation_total` on the prometheus metrics endpoint. 

IncrementValidationFailures is called directly by the reconciler and by the UpdateCassandra cluster event. 

The metric is initialised as part of the reconciler to ensure we capture changes to it

Required by https://github.com/sky-uk/cassandra-operator/issues/229


cassandra-operator/pkg/operator/reconciler.go
```diff
@@ -105,6 +108,9 @@ func (r *CassandraReconciler) Reconcile(request reconcile.Request) (reconcile.Re
 		return retryReconciliation("failed to lookup potential configMap", err)
 	}
 
+	// Initialise the failed validation metric just prior to validation.
+	r.metricsReporter.InitialiseFailedValidationMetric(desiredCassandra.Name, desiredCassandra.Namespace)
```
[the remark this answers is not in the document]
sebbonnet: should we move this method call to `determineChangesForCassandraDefinition` instead just before the Cassandra validation is invoked?

cassandra-operator/pkg/metrics/prometheus.go
```diff
@@ -28,6 +29,12 @@ func (t *clusterTopology) nodeCount() float64 {
 	return float64(len(t.nodesToRack))
 }
 
+// PrometheusMetricsInterface allows us to increment and initialise metrics via the PrometheusMetrics and Mocks
+type PrometheusMetricsInterface interface {
```
sebbonnet, line 33: Maybe call this `MetricsReporter` instead, so we don't have either Prometheus (the implementation) nor Interface in the name?

cassandra-operator/pkg/metrics/prometheus.go
```diff
@@ -28,6 +29,12 @@ func (t *clusterTopology) nodeCount() float64 {
...
+	IncrementFailedValidationMetric(clusterName string, clusterNamespace string)
```
sebbonnet, line 34: Should we use `*v1alpha1.Cassandra` as argument instead, to avoid mixing up the 2 args which look similar?
We could potentially expose the other functions (e.g. DeleteMetrics, UpdateMetrics, ...) in the interface and streamline them, so they all take a `*v1alpha1.Cassandra` as argument.

cassandra-operator/pkg/operator/controller_test.go
```diff
@@ -1001,3 +1028,20 @@ func (e *mockEventRecorder) PastEventf(object runtime.Object, timestamp metav1.T
...
+// force implementation of metrics.PrometheusMetricsInterface at compilation time
+var _ metrics.PrometheusMetricsInterface = &mockPrometheusMetrics{}
+
+func (m *mockPrometheusMetrics) IncrementFailedValidationMetric(clusterName string, clusterNamespace string) {
+	m.Called(clusterName, clusterNamespace)
+}
+
+func (m *mockPrometheusMetrics) InitialiseFailedValidationMetric(clusterName string, clusterNamespace string) {
+	m.Called(clusterName, clusterNamespace)
+}
```
sebbonnet, line 1047: newline missing

sebbonnet (changes requested):
Thanks for the contribution Chris!

Could you add some additional checks in the e2e tests so we can see the new metrics in actions? 
You'll find that we already check `InvalidChangeEvent` are reported in `invalid_modification_test`, so it would be good to ensure the new metric is reported there too.
I also think we should ensure the counter is initialised to `0` and so you could also add a check in the `creation_test` under `should report metrics for all clusters`.

If you're planning on releasing this change, then we'd also need a line in the CHANGELOG

cassandra-operator/pkg/metrics/prometheus.go
Bez625: I have changed this to `ClusterMetricsReporter` as `MetricsReporter` failed the lint check

cassandra-operator/pkg/operator/controller_test.go
Bez625: I have added this, but it didn't show up in lint and I can't see any references to it being a particular go coding style. 

Is this a coding convention that was agreed to as part of the operator? I'm wondering if we can add build time tests to enforce this going forward.

cassandra-operator/pkg/metrics/prometheus.go
Bez625: I agree with this and I have changed for the new functions. I looked at updating the existing functions and I was a bit apprehensive to do so as part of this PR as it is quite a big change.

Let me know if you feel it should be done as part of this PR and I'll update them

sebbonnet: I would rather do it now that we've introduced the interface. It's a low risk change and will give you the opportunity to view a bit more of the code  :)

cassandra-operator/pkg/operator/controller_test.go
sebbonnet: it's not a code convention, but more like a best practice as some unix tools would not cope well without it (wc is one example I can think of). github flags the missing new line in the changeset.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+## 0.82.2-alpha
```
sebbonnet, line 3: We should increment the minor version for new feature. The patch version should be incremented for bug fixes.
So here it should be 0.83.0-alpha

cassandra-operator/pkg/metrics/prometheus.go
```diff
@@ -141,6 +153,16 @@ func (m *PrometheusMetrics) UpdateMetrics(cluster *pkgcluster.Cluster) {
 	m.clustersMetrics.clusterSizeGauge.WithLabelValues(cluster.Name(), cluster.Namespace()).Set(clusterLastKnownTopology.nodeCount())
 }
 
+// IncrementFailedValidationMetric increment failed validation metrics metrics for the given cluster
+func (m *PrometheusMetrics) IncrementFailedValidationMetric(desiredCassandra *v1alpha1.Cassandra) {
```
sebbonnet, line 157: `desiredCassandra` can we call that simply `cassandra`, the "desired" part is only used when we reconcile the "desired" state based on the "actual" state

cassandra-operator/pkg/metrics/prometheus.go
```diff
@@ -141,6 +153,16 @@ func (m *PrometheusMetrics) UpdateMetrics(cluster *pkgcluster.Cluster) {
...
+	m.clustersMetrics.clusterValidationFailures.WithLabelValues(desiredCassandra.Name, desiredCassandra.Namespace).Inc()
+}
+
+// InitialiseFailedValidationMetric increment failed validation metrics metrics for the given cluster
+func (m *PrometheusMetrics) InitialiseFailedValidationMetric(desiredCassandra *v1alpha1.Cassandra) {
```
sebbonnet, line 162: same comment as above

cassandra-operator/pkg/operator/event/receiver.go
```diff
@@ -38,11 +38,11 @@ type OperatorEventReceiver struct {
 }
 
 // NewEventReceiver creates a new OperatorEventReceiver
-func NewEventReceiver(clusterAccessor cluster.Accessor, metricsPoller *metrics.PrometheusMetrics, eventRecorder record.EventRecorder) Receiver {
+func NewEventReceiver(clusterAccessor cluster.Accessor, metricsReporter *metrics.PrometheusMetrics, eventRecorder record.EventRecorder) Receiver {
```
sebbonnet, line 41: I would use the interface here once we expose the other functions

sebbonnet (commented):
Just a couple more points to address.
Let's use the ClusterMetricsReporter interface. It's a low risk change and will give you the opportunity to view a bit more of the code :)

sebbonnet (approved):

PR #234, opened 2020-11-11
Add metric getter for e2e-test 

Bez625:
This is a small change to the logic in the e2e tests to verify a metric has incremented as opposed to asserting on an absolute value.


cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
 	}
 }
 
+func GetMetric(metricGetterFunc func() (string, error), metricToFind string) (*int, error) {
```
alangibson01, line 185: I'd just return `int` and use a 0 return value when returning an error.

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
+
+	metricValue, err := parseMetricForValue(metrics, metricToFind)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse metrics with %s", err)
+	}
+
+	return metricValue, nil
+}
+
+func parseMetricForValue(metrics string, metricToFind string) (*int, error) {
```
alangibson01, line 204: As above.

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
+	lines := strings.Split(metrics, "\n")
+	for _, line := range lines {
+		if strings.Contains(line, metricToFind) && !strings.Contains(line, "#") {
```
alangibson01, line 207: You could simplify this by using `strings.HasPrefix` and adding a single space at the end of `metricToFind`.

alangibson01 (changes requested):
Just a few minor suggested changes.

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -182,41 +182,41 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
 
 	return metricValue, nil
 }
 
-func parseMetricForValue(metrics string, metricToFind string) (*int, error) {
+func parseMetricForValue(metrics string, metricToFind string) (int, error) {
 	lines := strings.Split(metrics, "\n")
 	for _, line := range lines {
-		if strings.Contains(line, metricToFind) && !strings.Contains(line, "#") {
+		if strings.HasPrefix(line, metricToFind) {
```
alangibson01, line 207: I think you should search for `metricToFind + " "` here - otherwise you could incorrectly match on `metricxyz` when searching for `metricxy`. (I think. Unless metrics are guaranteed to be lexically ordered, but I wouldn't bank on that.)

alangibson01 (dismissed):

cassandra-operator/test/e2e/cluster_status.go
Bez625: Ah i was unsure what you meant by the space. I believe metrics follow this kind of format:
```
cassandra_failed_validation_total{
```
So a space would not work here. I _could_ add `metricToFind + "{"`, but i'm not entirely sure we would have metrics names that are subsets of other metric names due to the convention of suffixing `_total`. 

Please let me know if you wish for me to add the curly brace.

alangibson01: Ahhh, good point. We can probably leave it as-is, to be honest.

alangibson01 (dismissed):

alangibson01 (approved):

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
 	}
 }
 
+func GetMetric(metricGetterFunc func() (string, error), metricToFind string) (int, error) {
+	// TODO: This should probably work on a timeout similarly to Eventually.
```
sebbonnet, line 186: can we remove this TODO?

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
+	metrics, err := metricGetterFunc()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get metrics with %s", err)
```
sebbonnet, line 189: %s should be replaced by %v or even %w since golang 1.13

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
+	}
+
+	if !(strings.Contains(metrics, metricToFind)){
+		return 0, fmt.Errorf("could not find metric %s", metricToFind)
+	}
+
+	metricValue, err := parseMetricForValue(metrics, metricToFind)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse metrics with %s", err)
```
sebbonnet, line 198: same comment about the `%s`

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -181,6 +182,43 @@ func OperatorMetrics(namespace string) func() (string, error) {
...
+	lines := strings.Split(metrics, "\n")
+	for _, line := range lines {
+		if strings.HasPrefix(line, metricToFind) {
+			lineSplit := strings.Split(line, " ")
+			if len(lineSplit) < 2 {
+				return 0, fmt.Errorf("failed to parse metric %s as got more values than expected in %s", metricToFind, line)
+			}
+			conv, err := strconv.Atoi(lineSplit[1])
+			if err != nil {
+				return 0, fmt.Errorf("failed to parse metric %s with err %s", metricToFind, err)
```
sebbonnet, line 214: same comment above the `%s` for error

PR #236, opened 2020-11-24
Update kind version to most recent that supports k8 1.12

Bez625:
Update kind version from 0.5.1 to 0.8.1 as that seems to be the current version available from package managers at the moment. 

There are quite a few breaking changes across the minor releases, but I believe they have been addressed in this commit.


cassandra-operator/Makefile
```diff
@@ -14,7 +14,7 @@ CASSANDRA_SNAPSHOT_IMAGE ?= $(TEST_REGISTRY)/cassandra-snapshot:v$(gitRev)
 DOMAIN ?= localhost
 IMAGE_TO_DEPLOY ?= $(dockerTestImage)
 NAMESPACE ?= test-cassandra-operator
-GINKGO_PARALLEL_OPTIONS ?= $(if $(GINKGO_NODES),-p -nodes=$(GINKGO_NODES),-p)
+GINKGO_PARALLEL_OPTIONS ?= $(if $(GINKGO_NODES),-p -nodes=$(GINKGO_NODES),)
```
alangibson01, line 17: What's the reason behind this change?

cassandra-operator/test/e2e/cluster_status.go
```diff
@@ -188,7 +188,7 @@ func GetMetric(metricGetterFunc func() (string, error), metricToFind string) (in
 		return 0, fmt.Errorf("failed to get metrics with %v", err)
 	}
 
-	if !(strings.Contains(metrics, metricToFind)){
+	if !(strings.Contains(metrics, metricToFind)) {
```
alangibson01, line 191:
```suggestion
	if !strings.Contains(metrics, metricToFind) {
```

I don't think the surrounding () are necessary in Go.

alangibson01 (changes requested):
One very minor change and one question.

cassandra-operator/test/e2e/cluster_status.go
Bez625: Ok this was existing code but happy to change it!

cassandra-operator/Makefile
Bez625: When I ran with `-p` and no `-nodes=` value locally the tests never seem to terminate. Remove `-p` if no GINKO_NODES seemed to be the logical choice as the tests worked fine in that case. 

If this is a surprise or a change in behaviour to how the testing has worked previously maybe we should catch up offline?

alangibson01: No, not a surprise at all. I found this as well when I was working on the operator, but never got round to fixing it!

alangibson01 (approved):

PR #237, opened 2020-12-02
Remove CPU resource limit from cassandra-sidecar

Bez625:
The current behaviour for resources on the cassandra-sidecar container is to match CPU limit to CPU requests. This leads to CPU throttling whenever the container needs to burst and this can be undesirable in some deployments. 

The intention is to remove this default behaviour in the short term and in the long term have it configurable via the cassandra resource object.


cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -521,7 +521,6 @@ var _ = Describe("creation of stateful sets", func() {
 
 			// then
 			Expect(sidecarContainer.Name).To(Equal("cassandra-sidecar"))
-			Expect(*sidecarContainer.Resources.Limits.Cpu()).To(Equal(resource.MustParse("100m")))
```
chuckydev, line 524: Ive just thought. Rather than removing the expectation, we should maybe verify there is not CPU Limit set at all

chuckydev (changes requested):

chuckydev (approved):

Issue #238, opened 2020-12-04
Provide configuration in cassandra resource for snapshot job


Bez625:
## Motivation
It is ideal to be able to specify the resource required by the cassandra-snapshot job that is generated as part of the snapshot process. This will allow users to be able to provision the necessary resources in their cluster namespace for the job.

## Requirements
It will be possible to set the cpu and memory requests for the snapshot job within the cassandra resource.

## Alternatives
We could hard code the values to specific levels and advertise them.

## Tasks/Feature Set
- [x] Write tests to ensure that the snapshot job is generated with specific resource requests
- [ ] ~~Ensure that sensible defaults are defined and used if omitted in the cassandra config.~~ See https://github.com/sky-uk/cassandra-operator/issues/238#issuecomment-754565719
- [x] Document the process on the wiki

Bez625:
We decided to remove the concept of default resources values for both snapshot and snapshot clean up and make both mandatory to be supplied on the cassandra spec. This will be a breaking change, but we believe that is acceptable for an alpha and we will make that clear in the changelog.

PR #239, opened 2020-12-08
Introduce default values for resources on snapshot job and snapshot cleanup job

Bez625:
required by https://github.com/sky-uk/cassandra-operator/issues/238

This is an initial PR to set default resources on the snapshot and cleanup job. The intention is to have further PRs that will build on this allowing those values to be configurable on the Cassandra Resource itself.


chuckydev:
Would you be able to modify the CHANGELOG then I would be ok to accept. I think this would be a release on its own as we dont have any immediate changes to make after this PR?

chuckydev (approved):

PR #240, opened 2020-12-10
Add configurable snapshot and snapshot-cleanup job resources

Bez625:
required by https://github.com/sky-uk/cassandra-operator/issues/238

* Add resources for Snapshot and Snapshot.RetentionPolicy within the Cassandra Resource Spec
* Apply resources to snapshot job and snapshot-cleanup job when those are created or modified.
* If neither are supplied we set the requirements to the previous default values (limits: 50Mi, requests: 100m, 50Mi).
* Testing for the above cases.


cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -277,25 +277,37 @@ func (c *Cluster) CreateSnapshotJob() *v1beta1.CronJob {
...
-			},
-		},
+		Name:      c.definition.SnapshotJobName(),
+		Image:     *c.definition.Spec.Snapshot.Image,
+		Command:   c.snapshotCommand(),
+		Resources: generateResources(c.definition.Spec.Snapshot.Resources),
+	}
+}
+
+func generateResources(clusterResources v1.ResourceRequirements) v1.ResourceRequirements {
```
sebbonnet, line 294: Default values for fields we expose in the Cassandra definition should all be set in the same place (via `v1alpha1helpers.SetDefaultsForCassandra`).
This is so the reconciler can compare both the current and desired cassandra based on the same default values. 
I think if you added a test that creates a Snapshot without resource requirements the reconciler would always find that the desired Cassandra is always different than the actual one as you only apply default to the actual one.
Also keeping the code in the same place would make it easier if we ever decide to write a mutating validation webhook that would set the defaults using that single entry point.

sebbonnet: Conversely the reconciler should apply the changes if you change the values of the Snapshot resource requirements.
I can't see any change in the snapshot state reconciliation `snapshotstate.go`, so I don't think this would work.
Do we have a test for that?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -191,6 +194,9 @@ type RetentionPolicy struct {
 	// Time limit for the cleanup command to run.
 	// +optional
 	CleanupTimeoutSeconds *int32 `json:"cleanupTimeoutSeconds,omitempty"`
+	// Resource requirements for the Snapshot Cleanup Job
+	// +optional
+	Resources coreV1.ResourceRequirements `json:"resources"`
```
sebbonnet, line 199: same comment about api conventions

sebbonnet (changes requested):
I don't think the reconciliation logic is going to work for snapshots when you change the resource requirements as the defaults are not applied to both desired and actual cassandra and there is no logic to reconcile the snapshot resources.
Also remember to add that change in the changelog - assuming you're going to release this :)

cassandra-operator/kubernetes-resources/cassandra-operator-deployment.yml
```diff
@@ -50,7 +50,7 @@ spec:
           value: $OPERATOR_IMAGE
         args: $OPERATOR_ARGS
         image: $OPERATOR_IMAGE
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always # For dev purposes as kind caches images and will not pull images with same commit tag
```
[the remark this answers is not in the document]
Bez625: This is something both @chuckydev and I found during development. We were trying to iterate on code locally and run tests using kind. As we were working on local changes the git tag (and therefore image tag) wasn't changing when the code was. 

I believe as per [the notes here](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) that it will still use caches for the image layers, so it is still fairly efficient to use this strategy. I'm happy to discuss this though as it is quite opinionated and potentially changes existing dev patterns.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -178,6 +178,9 @@ type Snapshot struct {
 	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
 	// +optional
 	RetentionPolicy *RetentionPolicy `json:"retentionPolicy,omitempty"`
+	// Resource requirements for the Snapshot job
+	// +optional
+	Resources coreV1.ResourceRequirements `json:"resources"`
```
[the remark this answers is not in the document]
Bez625: So i've taken a look at this and im a little confused. If I look at the Container struct within the [kubernetes api](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2258) they do NOT seem to be following this convention.

From what I can see `Resources` is both optional and a type (struct) that does not have a built in value of nil. Am I missing something here?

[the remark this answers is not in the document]
Bez625: The reason I'm questioning this is that it's kind of annoying as it means we can have a valid nil value for the resources defined on the cassandra snapshot spec, but then have to provide _something_ to the Container resources value.

cassandra-operator/kubernetes-resources/cassandra-operator-deployment.yml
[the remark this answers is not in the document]
sebbonnet: if you need to make a change you can always do another commit and rebuild your image or tag your image with latest.
I think using the default policy for pulling images promotes the use of immutable images which is a good thing to have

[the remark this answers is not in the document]
sebbonnet: I think it's wrong to change this just to accommodate your testing

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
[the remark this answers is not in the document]
sebbonnet: > So i've taken a look at this and im a little confused. If I look at the Container struct within the kubernetes api they do NOT seem to be following this convention.

:laughing: well it looks like they don't follow their own convention

> The reason I'm questioning this is that it's kind of annoying as it means we can have a valid nil value for the resources defined on the cassandra snapshot spec

You cannot have `nil` value with a struct - it needs to be a pointer.
Without using pointers you won't be able to tell whether the user request a cpu with value "0" or no cpu at all and therefore you don't know when you can apply the defaults.
Try this
```
			var noValue resource.Quantity
			zeroValue := resource.MustParse("0")
			Expect(noValue.Value()).To(Equal(zeroValue.Value()))

			var optionalValue *resource.Quantity
			// panics
			Expect(optionalValue.Value()).To(Equal(zeroValue.Value()))
```

cassandra-operator/Makefile
```diff
@@ -34,8 +34,15 @@ all: clean install check
 
 build: go-build check-style validate-licences validate-crds validate-codegen docker-image
 
+clean-test-namespace:
+	@echo "== delete all in ${NAMESPACE}"
+	kubectl --context ${KUBE_CONTEXT} -n ${NAMESPACE} delete all --all
+	kubectl --context ${KUBE_CONTEXT} -n ${NAMESPACE} delete cassandras --all
+
```
alangibson01, line 41: This feels a little dangerous to me, especially the `delete all` line. All it takes is a badly-defined env var and damaging things could happen. It'd be better to limit just to a certain set of fixed namespaces if possible.

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -399,6 +420,90 @@ var _ = Describe("Cassandra Helpers", func() {
...
+					coreV1.ResourceCPU:    resource.MustParse("100m"),
+				},
+				Requests: coreV1.ResourceList{
+					coreV1.ResourceCPU:    resource.MustParse("100m"),
+					coreV1.ResourceMemory: resource.MustParse("50Mi"),
+				},
+			}
+			Expect(SnapshotPropertiesUpdated(snapshot1, snapshot2)).To(BeFalse())
+			Expect(SnapshotPropertiesUpdated(snapshot2, snapshot1)).To(BeFalse())
+		})
```
alangibson01, line 463: I don't think ordering of properties in struct declarations is relevant. You can remove this test.

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra_test.go
```diff
@@ -462,5 +567,84 @@ var _ = Describe("Cassandra Helpers", func() {
...
+					coreV1.ResourceMemory: resource.MustParse("55Mi"),
+					coreV1.ResourceCPU:    resource.MustParse("0"),
+				},
+				Requests: coreV1.ResourceList{
+					coreV1.ResourceMemory: resource.MustParse("55Mi"),
+				},
+			}
+			Expect(SnapshotCleanupPropertiesUpdated(snapshot1, snapshot2)).To(BeFalse())
+			Expect(SnapshotCleanupPropertiesUpdated(snapshot2, snapshot1)).To(BeFalse())
+		})
```
alangibson01, line 608: As above.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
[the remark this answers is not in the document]
alangibson01: Is this conversation out-of-date now that resources are mandatory?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
alangibson01: As above.

alangibson01 (changes requested):

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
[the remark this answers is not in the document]
Bez625: Yeah I think so. We came full circle

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
Bez625: Yep outdated following the mandatory decision

cassandra-operator/Makefile
Bez625: Fair enough. Given this is a util method and not used by the travis/jenkins pipelines I will hardcode the namespace to the test namespace.

cassandra-operator/test/e2e/parallel/modification/external_modification_test.go
```diff
@@ -84,4 +105,157 @@ var _ = Context("External cluster modifications", func() {
...
+					coreV1.ResourceMemory: resource.MustParse("87Mi"),
+				},
+				Requests: coreV1.ResourceList{
+					coreV1.ResourceMemory: resource.MustParse("87Mi"),
+					coreV1.ResourceCPU:    resource.MustParse("287m"),
+				},
+			}
+			editedSnapshotJob := aCluster.CreateSnapshotJob()
+			editedSnapshotJob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Resources = editedSnapshotResources
+			TheSnapshotCronJobIsEdited(Namespace, clusterName, editedSnapshotJob)
```
sebbonnet, line 199: I think we should avoid calling functions from the production code such as `aCluster.CreateSnapshotJob()`
Can we do something similar to `TheClusterPodResourcesSpecAreChangedTo` where we mutate a copy of the server representation of the object?

cassandra-operator/test/e2e/parallel/modification/external_modification_test.go
```diff
@@ -84,4 +105,157 @@ var _ = Context("External cluster modifications", func() {
...
+					coreV1.ResourceMemory: resource.MustParse("87Mi"),
+				},
+				Requests: coreV1.ResourceList{
+					coreV1.ResourceMemory: resource.MustParse("87Mi"),
+					coreV1.ResourceCPU:    resource.MustParse("287m"),
+				},
+			}
+			editedSnapshotCleanupJob := aCluster.CreateSnapshotCleanupJob()
+			editedSnapshotCleanupJob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Resources = editedSnapshotCleanupResources
+			TheSnapshotCleanupCronJobIsEdited(Namespace, clusterName, editedSnapshotCleanupJob)
```
sebbonnet, line 213: same as above

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -909,6 +909,32 @@ var _ = Describe("creation of snapshot job", func() {
 		Expect(*securityContext.FSGroup).To(Equal(GroupID))
 		Expect(*securityContext.RunAsGroup).To(Equal(GroupID))
 	})
+	It("should create a cronjob with the specified memory and cpu resources", func() {
```
sebbonnet, line 912: Can we have a similar test for the snapshot cleanup job?

sebbonnet (commented):
some more comments on how to improve the tests

cassandra-operator/pkg/cluster/cluster.go
sebbonnet: this is fixed in more recent commits

sebbonnet (approved):
lgtm

cassandra-operator/test/e2e/parallel/modification/external_modification_test.go
sebbonnet: fixed in a more recent commit

alangibson01 (approved):

PR #250, opened 2021-01-29
Update code-gen to use vendor path

Bez625:
The code-gen functionality isnt working as expected. I have changed to output to the vendor path within the project directory and updating the source paths as appropriate

### DIFF
--- a/cassandra-operator/hack/update-codegen.sh
+++ b/cassandra-operator/hack/update-codegen.sh
@@ -9,24 +9,24 @@ projectDir="${scriptDir}/.."
 codeGenDir="${projectDir}/vendor/k8s.io/code-generator"
 basePackage="github.com/sky-uk/cassandra-operator/cassandra-operator"
 groupVersion="cassandra:v1alpha1"
+VENDOR_DIR="${projectDir}/vendor"
+GEN_FILE_PATH="/pkg/apis/cassandra/v1alpha1/zz_generated.deepcopy.go"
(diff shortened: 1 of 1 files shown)


Bez625:
Ok I now have a better idea of what is going on. As discussed there is an assumption in the gen code that expects your project source to be in your GOPATH, as was the convention prior to the introduction of go modules in 1.11. In that case I think we could probably change:
```
--output-base ${projectDir}/../../../..
```
to:
```
--output-base "`go env GOPATH`/src"
```
However, given we use go modules and therefore there is no requirement to have your development workspace within your GOPATH (see [here for detail](https://blog.golang.org/using-go-modules)), I dont think we can rely on this assumption any longer. 

In this PR I have made a change that uses the `/vendor` directory as a staging directory to generate the code and copy it over to the source code path. This is a bit of a hack as it takes advantage of the `go mod vendor` being called earlier in the `update-codegen.sh` script. 

It is worth noting that the reason we need a staging directory is that the code generation outputs to a path prefixed by the package structure - in this case `"github.com/sky-uk/cassandra-operator/cassandra-operator"`. I do not think we can change this behaviour.

An alternative approach could be to use a temp dir as a staging directory (similar to how it was done [in the verify script](https://github.com/sky-uk/cassandra-operator/blob/master/cassandra-operator/hack/verify-codegen.sh#L18)).

chuckydev (approved):

PR #251, opened 2021-02-01
Reduced creation CPU request

chuckydev:
Currently the sidecar modification tests are requesting unnecessary amounts of cpu. I have defaulted the requested to 0 and altered the modified to use 1m.

### DIFF
--- a/cassandra-operator/test/e2e/parallel/modification/sidecar_external_modifiation_test.go
+++ b/cassandra-operator/test/e2e/parallel/modification/sidecar_external_modifiation_test.go
@@ -106,7 +106,7 @@ var _ = Context("External sidecar statefulset modifications", func() {
 		//when
 		By("Modifying resource fields directly in the rack statefulset resource")
 		modifiedMemory := "60Mi"
-		modifiedCPU := "50m"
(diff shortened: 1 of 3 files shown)


cassandra-operator/test/e2e/parallel/modification/sidecar_external_modifiation_test.go
```diff
@@ -106,7 +106,7 @@ var _ = Context("External sidecar statefulset modifications", func() {
 		//when
 		By("Modifying resource fields directly in the rack statefulset resource")
 		modifiedMemory := "60Mi"
-		modifiedCPU := "50m"
```
Bez625, line 109: More for my benefit than anything - is this a fix? It looks like previously it modified `50m` to `50m` so wasnt really updating that value?

chuckydev: @sebbonnet requested that it should be reduced. It needed to be reduced a lot more, I think it went from 60m to 50m if my memory holds true

Bez625 (approved):

PR #252, opened 2021-02-03
Add env var block to cassandra pod spec

Bez625:
New feature to allow defining `env` block on the cassandra resource definition. This allows definitions of the form:
```yaml
pod:
  env:
    - name: SOME_VAR
      value: SOME_VAL
    - name: SOME_VAR_FROM_SECRET
      valueFrom:
        secretKeyRef:
          name: a-secret-ref
          key: key-inside-secret
```
We only allow environment variables with explicit values or from secrets. If using `valueFrom` then `secretKeyRef` is required and `fieldRef` or `configMapRef` will be ignored. Further details will be added to the wiki prior to release.


cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
```diff
@@ -110,10 +110,43 @@ type Pod struct {
...
+	// Env variables for the cassandra container
+	// +optional
+	Env *[]CassEnvVar `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,7,rep,name=env"`
 	// Sidecar container specification
 	Sidecar *Sidecar `json:"sidecar"`
 }
 
+// EnvVar represents an environment variable the operator will add to the Cassandra Container
+// This is almost identical to coreV1.EnvVar, but replaces the type of ValueFrom.
+type CassEnvVar struct {
```
chuckydev, line 122: Should we not use v1.EnvVar instead and validate instead of duplication?

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
```diff
@@ -542,6 +542,15 @@ var _ = Describe("ValidateCassandra", func() {
 				c.Spec.Snapshot.RetentionPolicy.Resources.Limits = coreV1.ResourceList{coreV1.ResourceMemory: resource.MustParse("60Mi")}
 			},
 		),
+		Entry(
```
chuckydev, line 545: Are we in need of some ValueFrom tests?

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -408,6 +408,50 @@ func (c *Cluster) objectMetadataWithOwner(name string, extraLabels ...string) me
...
+	} else {
+		envVar = v1.EnvVar{Name: cassEnvVar.Name, Value: cassEnvVar.Value}
+	}
+	envVars[i] = envVar
+
+}
+
+func (c *Cluster) createContainerEnvVars() []v1.EnvVar {
+
+	// If Spec.Pod.Env is ommitted we can simply create the auto populated env vars.
```
chuckydev, line 431: We could also escape early if Env was empty?

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -408,6 +408,50 @@ func (c *Cluster) objectMetadataWithOwner(name string, extraLabels ...string) me
...
+	if c.definition.Spec.Pod.Env == nil {
+		return []v1.EnvVar{getExtraClassPathVar()}
+	}
+	numOfReservedVariablesToBeAdded := 1 // Only EXTRA_CLASSPATH at this stage
+	specEnvLength := len(*c.definition.Spec.Pod.Env)
```
chuckydev, line 436: I am pro the constant above as its a bit of a magic number. This var however is a lil redundant and can be inlined

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -408,6 +408,50 @@ func (c *Cluster) objectMetadataWithOwner(name string, extraLabels ...string) me
...
+	containerEnvVars := make([]v1.EnvVar, specEnvLength+numOfReservedVariablesToBeAdded)
+
+	// Iterate over the CassEnvVars on the Cassandra Spec, skipping any reserved variables.
```
chuckydev, line 439: The code is descriptive enough. Shouldnt be a need for these comments (except maybe in the EXTRA_CLASSPATH case above)

cassandra-operator/pkg/cluster/cluster.go
```diff
@@ -408,6 +408,50 @@ func (c *Cluster) objectMetadataWithOwner(name string, extraLabels ...string) me
...
+	// We shouldnt have any reserved env vars at this stage as validation should have caught them.
+	i := 0
+	for _, cassEnvVar := range *c.definition.Spec.Pod.Env {
+		if !v1alpha1helpers.IsAReservedEnvVar(cassEnvVar.Name) {
+			addCassEnvVarToEnvVars(cassEnvVar, containerEnvVars, i)
+			i++
+		}
+	}
+
+	// Add env vars we populate. As there is only one simply add at i, if the list grows we can iterate.
```
chuckydev, line 449: As above

cassandra-operator/pkg/cluster/cluster_test.go
```diff
@@ -570,11 +570,47 @@ var _ = Describe("creation of stateful sets", func() {
 		})
 	})
 
-	It("should define environment variable for extra classpath in main container", func() {
-		cluster := ACluster(clusterDef)
+	Describe("environment variables", func() {
```
chuckydev, line 573: Empty / Nil state doesnt seem to be tested

cassandra-operator/test/e2e/parallel/creation/creation_test.go
```diff
@@ -116,8 +116,13 @@ func createClustersInParallel(multipleRacksCluster, emptyDirCluster *TestCluster
 					InitialDelaySeconds: ptr.Int32(CassandraInitialDelay),
 					PeriodSeconds:       ptr.Int32(CassandraReadinessPeriod),
 				},
-			},
-		}).IsDefined()
+				Env: &[]v1alpha1.CassEnvVar{
```
chuckydev, line 119: These do not seem to be validated anywhere in creation_test?

cassandra-operator/pkg/cluster/clusterstate.go
```diff
@@ -175,6 +177,44 @@ func (cass *currentClusterStateFinder) initContainerWithName(name string, statef
 	return nil, fmt.Errorf("no init container with name %s found in statefulset %s", name, statefulSet.Name)
 }
 
+func (cass *currentClusterStateFinder) buildEnvVarsFrom(envVars []corev1.EnvVar) *[]v1alpha1.CassEnvVar {
```
chuckydev, line 180: Does this need tested in clusterstate_test?

cassandra-operator/test/e2e/matchers.go
```diff
@@ -223,6 +227,7 @@ type ContainerExpectation struct {
 	ReadinessProbeSuccessThreshold int32
 	ReadinessProbePeriod           time.Duration
 	ReadinessProbeInitialDelay     time.Duration
+	CassEnvMap                     map[string]string
```
chuckydev, line 230: Just Env? Map is unnecessary as its type is defined here.

cassandra-operator/test/e2e/test_environment.go
```diff
@@ -45,6 +45,7 @@ var (
 	NodeTerminationDuration                 time.Duration
 	Namespace                               string
 	MaxCassandraNodesPerNamespace           int
+	CassEnvMap                              map[string]string
```
chuckydev, line 48: Just Env?

chuckydev (changes requested):

cassandra-operator/pkg/apis/cassandra/v1alpha1/helpers/cassandra.go
```diff
@@ -13,6 +13,13 @@ import (
 	"github.com/sky-uk/cassandra-operator/cassandra-operator/pkg/util/ptr"
 )
 
+// IsAReservedEnvVar is used to determine if an Environment Variable has been reserved.
+func IsAReservedEnvVar(envVar string) bool {
+	// String -> bool map should default to false for missing items.
```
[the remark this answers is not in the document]
Bez625: I can remove

cassandra-operator/test/e2e/parallel/reconciliation/reconciliation_test.go
```diff
@@ -1,19 +1,49 @@
-package modification
+package reconciliation
```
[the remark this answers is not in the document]
Bez625: I'm not 100% sure I understand the comment, are you saying that you think this is better in separate files? If so I'm happy to split in files similarly to how they were before, but in a different package

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
Bez625: I couldnt see any way of using v1.EnvVar without having to support fieldRefs and configMapFrom env vars. Happy if you can suggest a way we can

cassandra-operator/pkg/apis/cassandra/v1alpha1/validation/cassandra_test.go
Bez625: Yes, i'll add

cassandra-operator/pkg/cluster/cluster.go
Bez625: Agreed

cassandra-operator/pkg/cluster/cluster.go
Bez625: It makes the inline usage lengthy and a little hard to read, but agree it's only used once. I'm not sure on the Go conventions for readability over use once variables, but ill inline it

cassandra-operator/pkg/cluster/cluster.go
Bez625: I'll remove

cassandra-operator/pkg/cluster/cluster.go
Bez625: I'll remove

cassandra-operator/pkg/cluster/cluster_test.go
Bez625: It shouldnt be possible given we always add "EXTRA_CLASSPATH"

cassandra-operator/test/e2e/parallel/creation/creation_test.go
Bez625: They are validated on line 177?

cassandra-operator/pkg/cluster/clusterstate.go
Bez625: It is tested implicitly, but i can try and think of an explicit test to add. Maybe around it filtering out the EXTRA_CLASSPATH

cassandra-operator/test/e2e/matchers.go
Bez625: Agreed

cassandra-operator/test/e2e/test_environment.go
Bez625: Agreed

cassandra-operator/test/e2e/parallel/reconciliation/reconciliation_test.go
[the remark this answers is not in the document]
chuckydev: Happy with the move from external_modification -> reconciliation. 

Before we had sidecar_external_modiciation and external_modification. These have now been merged into a single test file.

cassandra-operator/pkg/cluster/cluster_test.go
chuckydev: Indeed. I mean on the CRD definition. Ie if no Env is defined, ensure EXTRA_CLASSPATH is still defined?

cassandra-operator/test/e2e/parallel/creation/creation_test.go
chuckydev: My bad. Was reading that as a passin, not a validation.

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
chuckydev: For reference:

https://pkg.go.dev/k8s.io/api/core/v1#EnvVar
https://pkg.go.dev/k8s.io/api/core/v1#EnvVarSource

Could we just not check for nil on everything except for SecretKeyRef as part of validation? This leaves room for simple expansion in the future if needed. The validation is restrictive as we have no purpose of testing it a the second I suppose. Ideally all options would eventually be supported

Bez625: I think when we have a concrete use case/requirement to support the other options we should expand the allowed options then. Given there is manipulation of the env vars between the cassandra resource and the stateful set i'd like to keep the set of possible config options small and well tested.

cassandra-operator/pkg/cluster/cluster_test.go
Bez625: Isn't that tested on line 583?

cassandra-operator/pkg/apis/cassandra/v1alpha1/types.go
chuckydev: Not suggeting supporting fields that are not required. Purely was trying to remove duplication from v1.EnvVar structures and validating.

However as raised elsewhere you make a good point that autogeneration will create documentation for these unsupported fields which is not ideal. Long term this can potentially return to being a v1.EnvVar, but for now happy to pass

chuckydev (approved):

PR #256, opened 2021-02-15
Add a utility method to deploy an operator without the cluster level CRD

Bez625:
Add a utility method to deploy an operator without the cluster level CRD. This can be useful if you are using the deploy operations of the repo in an existing cluster with a previously defined CRD.


cassandra-operator/hack/deploy.sh
```diff
@@ -57,10 +57,41 @@ function deployOperator() {
...
+            -e "s@\$OPERATOR_IMAGE@$operatorImage@g" \
+            -e "s@\$OPERATOR_ARGS@$operatorArgs@g" \
+            -e "s@\$INGRESS_HOST@$ingressHost@g" \
+            -e "s@\$VERSION@$imageVersion@g" \
+            ${resourcesDir}/${k8Resource} > ${tmpDir}/${k8Resource}
+        kubectl --context ${context} -n ${namespace} apply -f ${tmpDir}/${k8Resource}
+    done
+
+    waitForDeployment ${context} ${namespace} ${deployment}
+}
```
alangibson01, line 84: Could we do something a bit more parameterised on the existing function, rather than having another copy of it here?

alangibson01 (changes requested):

cassandra-operator/hack/deploy.sh
Bez625: Initially this is what I was going to do, but I feel in the longer term we may want to remove this option and have `deploy-operator` be used purely for the kind tests. If you feel the CRD-less option is worth keeping longer term then I'm happy to introduce a bit more code reuse with a view to it being a permanent option.

alangibson01: I think it might end up being useful, so maybe a bit of extra work now is worth the effort.

alangibson01 (approved):

Issue #38, opened 2018-12-12
Review Code Structure


alangibson01:
The `cassandra-operator/pkg/cluster/cluster.go` file is growing and currently stands at ~647~ 769 lines long. It's in danger of becoming too long and should be split up in whichever way seems most appropriate.

PR #258, opened 2021-02-15
Refactoring cluster.go

Bez625:
required by https://github.com/sky-uk/cassandra-operator/issues/38


chuckydev:
The changes do look good to me. Those methods that extended from the Cluster struct make sense as they do not directly modify the type themselves. Happy for others to take a look aswell

alangibson01 (approved):
LGTM too.

PR #262, opened 2021-06-21
Update operator to be compatible with 1.16 Kubernetes 

Bez625:
- Update Kind nodes to use a 1.16 k8
- Migrate all apis to those available in 1.16


test-kubernetes-cluster/recreate-kind-cluster.sh
```diff
@@ -173,7 +173,9 @@ nodes:
 EOF
 
 # node image officially supported for v0.8.1 - see https://github.com/kubernetes-sigs/kind/releases/tag/v0.8.0 for list of supported
```
eggsbenjamin, line 175: Just an aside but now that you're using k8s version greater than `1.12.x` it may be worth considering upgrading to the latest version of kind `0.11.1` as the desired version of k8s `1.16.x` is compatible with this. We have found it to be far more stable anyway.

Bez625: Yeah that's fair and 0.11.1 is what i've been using locally. I probably need to update in the comments and setup.

